PR #87, opened 2020-08-03
[experiment] Use ESBuild for both TypeScript and minification

developit:
I decided to give ESBuild a try. It doesn't support transpiling to ES5 yet, so it's not going to work for the `foo.legacy.js` output files being generated in this PR. But it is extraordinarily fast. Production builds take under 1s on my machine and the minified sizes are pretty close to what we have with Terser.

**Pros:**
- crazy fast. I'd estimate about 5x faster prod builds.
- removes both Sucrase and Terser as deps

**Cons:**
- It's a native binary, so it can't be inlined. \*
- Installation requires a direct download from the npm registry
- The project aims to replace Rollup, not just Babel/Terser. The usage here might be too limited.

\* There's a WASM version that works in Node, but it's slower than the native version (maybe that's fine?)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4043,6 +4043,12 @@
         "is-symbol": "^1.0.2"
       }
     },
+    "esbuild": {
+      "version": "0.6.14",
+      "resolved": "https://registry.npmjs.org/esbuild/-/esbuild-0.6.14.tgz",
+      "integrity": "sha512-coT5T5flVNaleZQAciBJCpkJ0xB+7TShhinPLUK0Zbz/yy385DKY3nGTNmHy6+oFARI3qnNlBRAbMYgxLN9/mA==",
+      "dev": true
+    },
     "escape-goat": {
       "version": "2.1.1",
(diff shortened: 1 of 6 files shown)


davezuko:
Is this still being considered, or is any help needed? I've been using esbuild in plenty of my personal projects and would love to see it incorporated into build tools like yours. In fact, it was one of the first things I searched for after discovering the project :).

developit:
@davezuko It needs to be rebased a bit before we can do much of a comparison, and then I'd like to run it against a few development + production apps to see the impact on build and HMR times. It's a nontrivial size increase for WMR and requires a separate file (but only one), but ESbuild is also very capable. The minification is not quite as effective as Terser's in the general case, but I recall the TypeScript support is better than Sucrase's.

Another option we have would be to release this as a plugin. It could inject a rollup plugin that transforms ts(x)/js(x) and minifies bundles, and set the `minify` option to false to disable WMR's own minification.

developit:
For those interested, here's the comparison of Terser VS ESBuild from back when I did this PR:

https://gist.github.com/developit/ba6bfe0ccdb65254c850b3d3f26e938d

PR #168, opened 2020-10-24
[DRAFT] - cjs module transforms

JoviDeCroock:
This is my initial attempt at implementing cjs-module-lexer, ... I couldn't get the tests to run properly on my windows machine so I'm going to defer this for when I'm at my MBP.

### DIFF
--- a/demo/public/cjs.js
+++ b/demo/public/cjs.js
@@ -1,0 +1,6 @@
+const { h } = require('preact');
+
+console.log(h);
+exports.Test = 'test2';
+
+module.exports = 'test';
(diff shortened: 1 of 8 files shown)


demo/public/cjs.js
```diff
@@ -0,0 +1,6 @@
+const { h } = require('preact');
+
+console.log(h);
+exports.Test = 'test2';
+
+module.exports = 'test';
```
ForsakenHarmony, line 6: Is this really something that should be allowed? Breaks in node

JoviDeCroock:
I'll try redo'ing this PR, it's gotten a bit outdated

PR #169, opened 2020-10-25
Experimental SSR plugin

developit:
This adds support for `wmr.config.js`, and uses it in the demo app to demonstrate an SSR middleware.

ToDo:

- [x] have the SSR plugin use the same `prerender()` export we support for prod prerendering
- [x] filter out verbose node module loader logs
- [ ] production SSR (maybe this is too different of a case?)

### DIFF
--- a/demo/public/index.tsx
+++ b/demo/public/index.tsx
@@ -34,7 +34,7 @@ export function App() {
 	);
 }
 
-if (typeof document !== 'undefined') {
+if (typeof window !== 'undefined') {
 	if (document.querySelector('.app')) {
 		hydrate(<App />, document.body);
 	} else {


JoviDeCroock:
Should we revive this into a custom plugin? Do we expect a production server running, if so will we have output targets that make it easy to publish i.e. person wants a `serverless` way to serve it or a normal deploy on a server?

Inviz:
I have an idea for project where I'd need to precompile some pages on demand, and currently I'm on the fence if I should go with preact/wmr for it... I'd really love to, but I wish there was a decent way to do it without doing crazy hacks. Don't wanna go with Rails :)

PR #199, opened 2020-12-02
fix(create-wmr): add "wmr" as devdep

lukeed:
Without this, the scaffolded template will only work via global `wmr` installations


marvinhagemeister (approved):

developit:
Looks like we also need to add `preact-iso`!

developit (approved):

PR #200, opened 2020-12-02
fix(types): swap `declare` order for `*.module.css` files

lukeed:
The previous order matched `*.css` declarations for a `styles.module.css` file because it comes first in the types. This meant that all `className={ styles.home }` failed to typecheck, since `styles` was matched as a string.

This change allows `styles.modules.css` to be treated as a `Mapping` type, where as `styles.css` is treated as a `string` when imported.

### DIFF
--- a/types.d.ts
+++ b/types.d.ts
@@ -41,17 +41,17 @@ declare interface NodeModule {
 }
 declare var module: NodeModule;
 
-declare module '*.css' { const url: string; export default url; }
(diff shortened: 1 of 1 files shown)


developit:
hah! I thought the precedence was the other way around.

developit (approved):

marvinhagemeister (approved):

PR #204, opened 2020-12-03
Move server address log in listen callback

### LINKED ISSUE #202: Error on starting server
This seems to only happen on dev and serve, builds (with prerender) are unaffected.

```
$ DEBUG=1 npm run start

> start
> wmr

TypeError: Cannot read property 'address' of null
    at c7e (/home/intrnl/Documents/projects/public/first-wmr-project/node_modules/wmr/wmr.cjs:2:2713816)
    at u7e (/home/intrnl/Documents/projects/public/first-wmr-project/node_modules/wmr/wmr.cjs:2:2714675)
npm ERR! code 1
npm ERR! path /home/intrnl/Documents/projects/public/first-wmr-project
npm ERR! command failed
npm ERR! command sh -c wmr
```

Checked it and it seems to fail around [here](https://github.com/preactjs/wmr/blob/0b368e7/src/lib/net-utils.js#L49)

- node 15.3.0
- npm 7.0.15
- OpenSUSE Tumbleweed 20201130-0 (5.9.11-1-default)

intrnl:
Moves the logging for server addresses in the listen callback, since we can't always assume that the `listening` event has already been emitted when we're retrieving the addresses.

Closes #202

### DIFF
--- a/src/start.js
+++ b/src/start.js
@@ -61,7 +61,8 @@ export default async function start(options = {}) {
 	}
 
 	const app = await server(options);
-	app.listen(options.port, options.host);
-	const addresses = getServerAddresses(app.server.address(), { https: app.http2 });
-	process.stdout.write(kl.cyan(`Listening on ${addresses}`) + '\n');
+	app.listen(options.port, options.host, () => {
+		const addresses = getServerAddresses(app.server.address(), { https: app.http2 });
+		process.stdout.write(kl.cyan(`Listening on ${addresses}`) + '\n');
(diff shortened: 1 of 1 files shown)


marvinhagemeister (approved):
Ohh that's much better! Makes a lot of sense to put it there instead. Thank you so much for the PR 🙌

intrnl:
By the looks of it, it seems that `serve.js` is also doing the same thing, sending a follow up :+1:

PR #210, opened 2020-12-03
chore: update readme `wmr.config.js` snippet

lukeed:
Added individual config functions so that that variant is more visible. Although it's mentioned in the paragraph above the code sample, it's still easy to gloss over.

Also changed the file annotation to `wmr.config.mjs` because ESM inside `.js` only works by default when `"type": "module"` is defined. Since that's not set on the `create-wmr` template, safe to assume `wmr.config.js` + ESM won't work for most out of the box.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,10 +117,10 @@ function App() {
 ## Configuration and plugins
 
 WMR supports a `wmr.config.js` (or `wmr.config.mjs`) configuration file.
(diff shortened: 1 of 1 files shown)


marvinhagemeister (approved):
Love this! Much easier to see this and not miss it 👍

lukeed:
I missed it 😅 And I even had vague recollection that it existed.

PR #213, opened 2020-12-03
feat: add `publicPath` config key

lukeed:
There was `publicPath: '/'` in a few places throughout the plugin instantiations. Grouped them & exposed definition to outside world... still defaults to `/` though.

This allows a dev to define/customize where their assets will live in production; for example:

```js
// wmr.config.mjs
  config.publicPath = config.prod ? 'https://cdn.example.com/' : '/';
}
```

In development, nothing changes (with this config), but in production, the HTML attachments include the prefix, as do any other external asset links (like the `.jpg` file in demo)

> **Note:** I did not add this as a CLI option. Will leave that decision up to others, but I didn't want to start us down the same PreactCLI track of having too many options.

---

***TODO***

- [x] add test(s)
- [x] fix `style()` helper, which is misbehaving

I've spent some time poking at the `style()` already to no avail. I think it might be something Rollup itself is doing, but production output looks like this:

```js
($w_s$("https:/cdn.example.com/assets/Calendar.75329e7d.css"),/* ... */)
```

Notice the single `/` in `https:/` 🤔 
Values coming out of `public-path-plugin.js` are correct though, so something is re-resolving it incorrectly.

_EDIT: Narrowed down the `https:/` issue to `optimize-graph-plugin`_


src/plugins/optimize-graph-plugin.js
```diff
@@ -373,14 +390,15 @@ function hoistTransitiveImports(graph) {
...
-				preloads.push(...css.map(f => `${meta.styleLoadFn}(${JSON.stringify(posix.join(graph.publicPath, f))})`));
+				preloads.push(...css.map(f => `${meta.styleLoadFn}(${toImport(graph.publicPath, f, true)})`));
 			}
 
 			const js = deps.js.get(spec);
 			if (js && js.length) {
 				if (DEBUG) console.log(`Preloading JS for import(${spec}): ${js}`);
 				preloads.push(
 					...js.map(f => {
+						// TODO: should have `toImport(...)` somewhere here
```
lukeed, line 401: TBH I'm not sure what's really going on with the next line. But in any event, its `graph.publicPath` isn't yet accounting for the fact that `publicPath` may be an external address.

I couldn't figure out how to reach this block of code, so couldn't confidently update this.

src/plugins/optimize-graph-plugin.js
```diff
@@ -176,15 +176,31 @@ class ChunkGraph {
...
 
+/**
+ * @param {string} publicPath
+ * @param {string} filename
+ * @param {boolean} [toJSON]
+ */
+function toImport(publicPath, filename, toJSON) {
+	let value = posix.join(publicPath, filename);
+
+	if (/^(https?:)?\/\//.test(publicPath)) {
```
lukeed, line 195: There's (amazingly) no `src/utils` directory yet, but extracting this logic (same as `toImport` inside optimize) is a good candidate to start it off.

developit: lol - yeah I've waffled a lot on a utils directory/module. This plugin probably warrants moving into its own repo at some point, and I didn't want to couple it to WMR's codebase as much as possible.

(agreed though)

src/bundler.js
```diff
@@ -71,6 +73,10 @@ export async function bundleProd({
 	cwd = cwd || '';
 	root = root || cwd;
 
+	if (!publicPath.endsWith('/')) {
+		publicPath += '/';
+	}
```
developit, line 78: The default `/` and trailing slash enforcement might be better in `src/lib/normalize-options.js` so that they're uniformly applied to all modes.

Edit: looks like you already have the default value there, so just the trailing. It can fall back to `/`.

src/plugins/optimize-graph-plugin.js
lukeed: Yup – figured that was the case knowing you 😆 Left it this way with mental note to come back if/when commons are extracted.

lukeed:
Aside from adding a test-fixture that just proves the build will succeed, I've no idea how properly test this. Doesn't look like there's existing pattern to follow

src/plugins/optimize-graph-plugin.js
developit: Ah yeah - I had completely ignored `publicPath` being a URL here, and assumed it was a `pathname`. We could switch from using `posix.join()` to using something like:

```js
function relative(filename, publicPath) {
  let href = new URL(publicPath, 'https://f.example.com');
  let url = new URL(filename, href);
  if (href.origin === 'https://f.example.com') return url.pathname + url.search;
  return url.href;
}
```

developit:
@lukeed closest would be [this test](https://github.com/preactjs/wmr/blob/6b25180b870aab137bb29f7f9cc8295d9f0067c3/test/production.test.js#L126). Basically needs a fixtures directory that sets the publicPath to something, then to either assert over the generated JS (looking for `$w_s$(`) or load the result in Chrome and check for CSS having been applied.

src/plugins/optimize-graph-plugin.js
lukeed: You're suggesting this helper to take the place of the full `toImport` util? 
It _looks_ like it should be doing the same thing but I might be overlooking something.

developit: Could make it always JSON-stringify, I don't think there are any cases left where it needs to skip that.

developit: oh, wow I literally hadn't scrolled down to see that you'd already written this haha. I actually prefer your explicit protocol check over my origin weirdness.

FWIW `posix.join()` would still mangle the protocol in `publicPath` since it collapses adjacent slashes.

```js
/**
 * @param {string} publicPath
 * @param {string} filename
 */
function toImport(publicPath, filename, toJSON) {
	if (/^(https?:)?\/\//.test(publicPath)) {
        return JSON.stringify(new URL(filename, publicPath).href);
	} else {
		return JSON.stringify(posix.join(publicPath, filename));
	}
}
```

src/plugins/optimize-graph-plugin.js
```diff
@@ -176,15 +176,31 @@ class ChunkGraph {
 				let rep = typeof newUrl === 'function' ? newUrl(url, fn, quote) : newUrl;
 				if (rep === false) return 'null';
 				if (typeof rep === 'string') {
-					if (publicPath) rep = posix.join(publicPath, rep);
-					rep = JSON.stringify(rep);
+					rep = publicPath ? toImport(publicPath, rep, true) : JSON.stringify(rep);
```
developit, line 179: As per [other comment](https://github.com/preactjs/wmr/pull/213/files#r535578173):

```suggestion
					rep = JSON.stringify(publicPath ? toImport(publicPath, rep) : rep);
```

src/plugins/optimize-graph-plugin.js
lukeed: The `toJSON` is a parameter because of this TODO, which needs a regular string to figure out if/what addl processing needs to be done : https://github.com/preactjs/wmr/pull/213/files#diff-6b6c2718597d2a9b4a4334e7c5f6208ffe00946ac37ded1a170cdaa6cfc7dd06R401

Your most recent snippet fails the `//example.com/asset.jpg` case. That's why my `toImport` version deals with that `isFull` and `root` juggling act. Using `new URL` fails when the base does not include a fully qualified protocol.

src/plugins/optimize-graph-plugin.js
developit: ah - these are Rollup-generated `import()` statements and already have the public path prepended.

lukeed: Ok, so they're safe to ignore?

src/plugins/optimize-graph-plugin.js
developit: Ah right I always forget Node doesn't support protocol-relative URLs (boo). So maybe this combo:

```js
/**
 * @param {string} publicPath
 * @param {string} filename
 */
function toImport(publicPath, filename) {
	const FAKE = 'https://a.a';
	const href = new URL(publicPath, FAKE);
	const url = new URL(filename, href);
	return JSON.stringify(href.origin === FAKE ? url.pathname + url.search : url.href);
}
```

Screen Shot 2020-12-03 at 3 57 36 PM

src/plugins/optimize-graph-plugin.js
lukeed: I'm still concerned about this. Even if they're already `publicPath`d by Rollup, we're still doing this, which is causing all the issues with external/protocol'd `publicPath` values:

```js
posix.join(graph.publicPath, f)
```

src/plugins/optimize-graph-plugin.js
lukeed: Treating `//example.com` the same as `https://example.com` is a mistake.

While we 100% hope/want people to be using HTTPS, that's not what they entered. The `publicPath` should be respected as is. Use of `//example.com` may be intentional so that websites always resolve using the _current_ protocol. 

(This happens.)

lukeed: With the last commit, these are the `toImport` outputs using your test values. They are expected:

```js
toImport('/', './foo/bar.js'); 
//=> '/foo/bar.js'

toImport('//example.com', './foo/bar.js')
//=> '//example.com/foo/bar.js'

toImport('//example.com', '../foo/bar.js')
//=> '//example.com/foo/bar.js'

toImport('//example.com', 'foo/bar.js')
//=> '//example.com/foo/bar.js'

toImport('http://example.com', 'foo/bar.js')
//=> 'http://example.com/foo/bar.js'

toImport('https://example.com', 'foo/bar.js')
//=> 'https://example.com/foo/bar.js'
```

src/bundler.js
lukeed: The `endsWith` check is to prevent the common webpack config issue:

```js
// wrong, but common mistake
config.publicPath = 'https://cdn.example.com';
```

src/plugins/optimize-graph-plugin.js
developit: Ah, yeah good point. One option might be to just skip the conversion to relative if publicPath contains `//` ?

developit (approved):
This looks like it's pretty much ready-to-go aside from the JS subresource prefetching thing. I think the case that triggers that is when dynamically importing a module that contains dynamic imports:

```js
// index.js
import('./foo.js');

// foo.js
import('./bar.js');  // gets hoisted into index.js as `import('/chunks/bar.js')`

// bar.js
console.log('blah');
```

developit:
I approved but I definitely want to at least do a gut-check on the nested dynamic import issue before release.

developit (approved):

PR #214, opened 2020-12-03
feat: add `minify` config

lukeed:
- enabled by default (no change)
- disable via `--no-minify` or `--minify=0` or `--minify=false`
- disable via `config.minify = false`

---

May be good to have this to debug what's going on w/ production. (Not everyone can follow Terser output 😉)
Am roughly doing this to debug what's going on with #213 

***TODO***

* Rename this to `--compress`?
    _Would be common term between commands, but `build`'s meaning would be different._

* Hide this from the CLI help text?
    _Could still attach it to CLI, but not advertise it in `wmr build -h` output._

* Do we even want this?
    _Didn't discuss this with anyone, sorry~!_

### DIFF
--- a/src/bundler.js
+++ b/src/bundler.js
@@ -35,6 +35,7 @@ const pathToPosix = p => p.split(sep).join(posix.sep);
  * @property {string} [publicDir = '']
  * @property {string} [out = '.cache']
  * @property {boolean} [sourcemap]
+ * @property {boolean} [minify]
  * @property {Record<string, string>} [aliases] module aliases
  * @property {boolean} [profile] Enable bundler performance profiling
  * @property {Record<string, string>} [env]
@@ -66,6 +67,7 @@ export async function bundleProd({
 	env = {},
 	plugins,
 	output,
+	minify = true,
 	npmChunks = false
 }) {
 	cwd = cwd || '';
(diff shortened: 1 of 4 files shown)


developit:
IIRC we already have `--no-compress` for turning off dynamic gzip response compression, so `minify` is likely best here.

src/bundler.js
```diff
@@ -136,7 +138,7 @@ export async function bundleProd({
 			return fileName;
 		},
 		hoistTransitiveImports: true,
-		plugins: [terser({ compress: true, sourcemap })],
+		plugins: [minify && terser({ compress: true, sourcemap })],
```
developit, line 141: A thought: we could allow `minify` to be an object in addition to a Boolean, with the format:

```js
{ terser: Partial<TerserConfig>, css: Partial<CSSNanoConfig> }
```

developit (approved):
Beauty! one Q but otherwise this is a nice add.

lukeed:
Right, that touches on my first "TODO" point. Felt it odd to share a name but give it different meaning 1/3 times.

src/bundler.js
lukeed: Personally that feels a bit too deep, but happy to add it if desired. 

In my stuff, I add separate `cssnano` and/or `terser` config blocks so that they're separately exportable. But then again, I always use the `@pwa/cli` (and now `freshie`) format where every config key is top-level.

developit:
I'm happy with `--minify`, and I actually like the idea of hiding it from the CLI helptext. We can treat it more like a config option for `wmr.config.js`, especially if we (later?) add the object form.

developit (approved):

marvinhagemeister (approved):

PR #215, opened 2020-12-03
chore: purge unused plugins

lukeed:
I hate to do it, but these two have no references anywhere. There's a lot going on the repo so might be best to trim away the unnecessary bits.

There's also the `unpkg` plugin (and one other, but I forget) that has a commented-out import. I left those.


JoviDeCroock (approved):
Good spot!

developit:
Good catch on the watcher plugin. The dynamic import names I actually think we might need - I don't know why it wasn't being used...

lukeed:
Should I revert that file?

developit:
For now that's probably best - we can delete it once we've confirmed, but it wasn't supposed to be dead code :P

lukeed:
Restored via force push

developit (approved):
LGTM

PR #216, opened 2020-12-03
chore: replace `tiny-glob` with `totalist` for inputs

lukeed:
There was only `tiny-glob` usage inside `bundler.js` to collect HTML input files (updated) and inside `watch-plugin.js`, which was unused.

No need to bring in the `tiny-glob` dependency chain when just crawling for a single extension.

(~1.5kb => 200b gzip)

### DIFF
--- a/src/bundler.js
+++ b/src/bundler.js
@@ -10,7 +10,7 @@ import npmPlugin from './plugins/npm-plugin/index.js';
 import publicPathPlugin from './plugins/public-path-plugin.js';
 import minifyCssPlugin from './plugins/minify-css-plugin.js';
 import htmlEntriesPlugin from './plugins/html-entries-plugin.js';
-import { totalist } from 'totalist';
+import totalist from 'totalist';
 import aliasesPlugin from './plugins/aliases-plugin.js';
 import processGlobalPlugin from './plugins/process-global-plugin.js';
 import urlPlugin from './plugins/url-plugin.js';
--- a/src/bundler.js
+++ b/src/bundler.js
@@ -10,7 +10,7 @@ import npmPlugin from './plugins/npm-plugin/index.js';
 import publicPathPlugin from './plugins/public-path-plugin.js';
 import minifyCssPlugin from './plugins/minify-css-plugin.js';
 import htmlEntriesPlugin from './plugins/html-entries-plugin.js';
-import glob from 'tiny-glob';
+import { totalist } from 'totalist';
 import aliasesPlugin from './plugins/aliases-plugin.js';
 import processGlobalPlugin from './plugins/process-global-plugin.js';
 import urlPlugin from './plugins/url-plugin.js';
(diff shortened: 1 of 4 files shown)


JoviDeCroock (approved):
Nice find :shipit:

src/bundler.js
```diff
@@ -71,15 +71,15 @@ export async function bundleProd({
...
-
 	// note: we intentionally pass these to Rollup as posix paths
 	const ignore = /^\.\/(node_modules|dist|build)\//;
-	const input = htmlFiles.map(p => './' + pathToPosix(relative('.', p))).filter(p => !ignore.test(p));
+	/** @type {string[]} */ const input = [];
+
+	await totalist(cwd, (rel, abs) => {
+		if (ignore.test(abs)) return;
+		if (!/\.html?/.test(rel)) return;
+		input.push('./' + pathToPosix(relative('.', abs)));
```
lukeed, line 81: should this be relative from `root` (aka, true value for `cwd`)?
```suggestion
		input.push('./' + pathToPosix(relative(root, abs)));
```

developit: @lukeed yes I think that's correct. The `input` should look like `['./public/index.html']`.

lukeed: Yes it looks like that either way. The question is more like do we trust `root` or not

developit: I usually check by running in the `public` vs `nopublic` fixtures:

```sh
# with ./public
node --experimental-modules src/cli.js build --cwd test/fixtures/simple

# no ./public
node --experimental-modules src/cli.js build --cwd test/fixtures/import-relative
```

developit: `root` is definitely the right way to go here because it's the CWD for rollup, to which `input` has to be relative 👍

lukeed: Both `'.'` and `root` work identically with the fixtures.

(However, unrelated, `test/fixtures/import-relative` cannot be rebuilt unless its `dist` is deleted)

developit: heh - yeah we should probably gitignore `dist` within the fixtures directory. Fixtures get copied to a temp dir, but we don't filter out `dist` because they're not really supposed to be run directly like I recommended :P

developit:
Looks like the build error on 12.8 is because `totalist` exports an ESM package with `{ totalist }` but a CJS package with `module.exports=totalist`.

lukeed:
Hmm... must be something else going on – it's a named `totalist` export for both: https://unpkg.com/totalist@2.0.0/dist/index.js

Might be because `sirv` still uses `totalist@^1.0` which is `module.exports/export default`, but Rollup should be able to differentiate.

lukeed:
Guess that was it /shrug

lukeed:
Without the last commit, the built output included `totalist` twice because `sirv` relies on `totalist/sync` (for synchronous, non-`dev` startup). I know `async` operations are preferable, but I think this should be fine?

developit:
I'd prefer the duplication in this case since it's potentially a large directory tree.

lukeed:
This should be ready to go. Please squash this bad boy

PR #217, opened 2020-12-03
chore: add preact-iso

JoviDeCroock:
We can still decide to move this to a seperate repo but this way people have exposure to how the code looks already


developit (approved):
Nice!

developit (approved):

PR #219, opened 2020-12-03
chore: add missing imports to preact-iso code snippets

EmilTholin:


developit (approved):

developit:
Nice! I'm also going to a blurb about importing everything from `"preact-iso"`, since it's code-splitted by default anyway.

Issue #182, opened 2020-11-20
Reload/HMR not working w/11ty


Snugug:
## Expected
Changing the HTML or JS should cause either a HMR or a hard reload

## Actual
11ty moves the file over as expected, or compiles (as needed) but `wmr` doesn't pick up the changes. Doing a manual reload works as expected. I've also tested w/just WMR, no 11ty in the pipeline, and it doesn't appear to be working there, either.

## System
* MacOS 10.15.7
* Shell: zsh
* Node: 12.19.0
* WMR: 0.0.12

Here's a minimal test-case:

**.eleventy.js**
```js
module.exports = function (eleventy) {
  eleventy.addPassthroughCopy('js');

  return {
    dir: {
      input: 'views',
      output: 'public',
      includes: '../templates/includes',
      layouts: '../templates/layouts',
      data: '../data',
    },
    dataTemplateEngine: 'njk',
    markdownTemplateEngine: 'njk',
    htmlTemplateEngine: 'njk',
    templateEngineOverride: 'njk',
  };
};
```

**views/index.html**
```html
<!DOCTYPE html>
<html lang="en">
<head>
  <meta charset="UTF-8">
  <meta name="viewport" content="width=device-width, initial-scale=1.0">
  <title>WMR Test</title>
  <script type="module" src="/js/index.js"></script>
  <style>
    body {
      background: blue;
    }
  </style>
</head>
<body>
  <h1>Hello World</h1>
</body>
</html>
```

**js/index.js**
```js
console.log('Hello World');
```

**package.json**
```json
{
  "name": "wmr-11ty",
  "version": "1.0.0",
  "scripts": {
    "11ty:watch": "eleventy --watch",
    "wmr": "wmr",
    "start": "run-p 11ty:watch wmr"
  },
  "devDependencies": {
    "@11ty/eleventy": "^0.11.1",
    "npm-run-all": "^4.1.5",
    "wmr": "0.0.12"
  }
}
```

Snugug:
Hey, sorry to comment on a closed thread, but now that 1.1 is released I tried this again, and it still isn't working. I'm even able to remove 11ty from the stack, just running WMR from the root of a project and editing an `index.html` or `style.css` (or even a root, as opposed to submodule, `main.js` included as a module in my HTML) simply isn't reloading the site. If I do a manual reload my changes are served correctly, but no automatic reloading or HMR. Is there anything else I can share that'll help y'all debug this? Here's the full HTML that I've got:

```html
[...]

PR #220, opened 2020-12-03
fix hmr for modules not referencing module.hot

JoviDeCroock:
Fixes: https://github.com/preactjs/wmr/issues/182

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
 	"bin": "wmr.cjs",
 	"type": "module",
 	"scripts": {
-		"demo": "cd demo && node --experimental-modules ../src/cli.js --no-optimize --no-compress 2>&1",
+		"demo": "cd demo && node --experimental-modules ../src/cli.js 2>&1",
 		"demo:prod": "cd demo && node --experimental-modules ../src/cli.js build --prerender 2>&1",
 		"demo:serve": "cd demo && node --experimental-modules ../src/cli.js serve",
 		"dev": "PROFILE=true nodemon -w src --exec \"npm run -s demo\"",
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
 	"bin": "wmr.cjs",
 	"type": "module",
 	"scripts": {
-		"demo": "cd demo && node --experimental-modules ../src/cli.js 2>&1",
+		"demo": "cd demo && node --experimental-modules ../src/cli.js --no-optimize --no-compress 2>&1",
 		"demo:prod": "cd demo && node --experimental-modules ../src/cli.js build --prerender 2>&1",
 		"demo:serve": "cd demo && node --experimental-modules ../src/cli.js serve",
 		"dev": "PROFILE=true nodemon -w src --exec \"npm run -s demo\"",
--- a/src/plugins/wmr/plugin.js
+++ b/src/plugins/wmr/plugin.js
@@ -53,7 +53,7 @@ export default function wmrPlugin({ hot = true } = {}) {
 			}
 			return null;
 		},
-		transform(code, id, ...args) {
+		transform(code, id) {
 			const ch = id[0];
 			if (ch === '\0' || ch === '\b' || !/\.[tj]sx?$/.test(id)) return;
 			let hasHot = /(import\.meta\.hot|\$IMPORT_META_HOT\$)/.test(code);
--- a/src/plugins/wmr/plugin.js
+++ b/src/plugins/wmr/plugin.js
@@ -53,7 +53,7 @@ export default function wmrPlugin({ hot = true } = {}) {
 			}
 			return null;
 		},
-		transform(code, id) {
+		transform(code, id, ...args) {
 			const ch = id[0];
 			if (ch === '\0' || ch === '\b' || !/\.[tj]sx?$/.test(id)) return;
 			let hasHot = /(import\.meta\.hot|\$IMPORT_META_HOT\$)/.test(code);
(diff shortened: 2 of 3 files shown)


src/plugins/wmr/plugin.js
```diff
@@ -75,15 +75,16 @@ export default function wmrPlugin({ hot = true } = {}) {
...
 			const s = new MagicString(code, {
 				filename: id,
 				// Typings from MagicString are wrong, see: https://github.com/Rich-Harris/magic-string/pull/182
 				// @ts-ignore
 				indentExclusionRanges: undefined
 			});
-			if (hot) {
+
+			if (!hasHot) {
+				s.prepend(`import { createHotContext as $w_h$ } from 'wmr'; $w_h$(import.meta.url);`);
```
developit, line 86:
```suggestion
				s.append(`\nimport { createHotContext as $w_h$ } from 'wmr'; $w_h$(import.meta.url);`);
```

src/plugins/wmr/plugin.js
```diff
@@ -75,15 +77,16 @@ export default function wmrPlugin({ hot = true } = {}) {
...
 				// Typings from MagicString are wrong, see: https://github.com/Rich-Harris/magic-string/pull/182
 				// @ts-ignore
 				indentExclusionRanges: undefined
 			});
-			if (hot) {
+
+			if (!hasHot && !isProduction) {
+				s.append(`\nimport { createHotContext as $w_h$ } from 'wmr'; $w_h$(import.meta.url);`);
+			} else if (hot) {
 				s.append(after);
```
developit, line 90:
```suggestion
			if (hot) {
				if (!hasHot) {
					s.append(`\nimport { createHotContext as $w_h$ } from 'wmr'; $w_h$(import.meta.url);`);
				} else {
					s.append(after);
```

package.json
```diff
@@ -4,7 +4,7 @@
 	"bin": "wmr.cjs",
 	"type": "module",
 	"scripts": {
-		"demo": "cd demo && node --experimental-modules ../src/cli.js 2>&1",
+		"demo": "cd demo && node --experimental-modules ../src/cli.js --no-optimize --no-compress 2>&1",
```
developit: can we keep these? they just control gzip and node_modules optimization

src/plugins/wmr/plugin.js
```diff
@@ -65,33 +66,41 @@ export default function wmrPlugin({ hot = true } = {}) {
 				before += `const module={${hot ? 'hot:import.meta.hot' : ''}};\n`;
 			}
 
+			const hasExport = code.match(/\bexport\b/);
```
developit, line 69: We can reuse the ESM_KEYWORDS regex from [here](https://github.com/preactjs/wmr/blob/7ede32763b18cd56b29a10cf79e884469c3a052d/src/plugins/fast-cjs-plugin.js#L3):

```suggestion
			const hasExport = code.match(/(\bimport\s*(\{|\s['"\w_$])|[\s;]export(\s+(default|const|var|let)[^\w$]|\s*\{))/);
```

src/plugins/wmr/plugin.js
```diff
@@ -65,33 +66,41 @@ export default function wmrPlugin({ hot = true } = {}) {
...
-			if (code.match(/html`[^`]*<([a-zA-Z][a-zA-Z0-9.:-]*|\$\{.+?\})[^>]*>/) && code.match(/\bexport\b/)) {
+			if (code.match(/html`[^`]*<([a-zA-Z][a-zA-Z0-9.:-]*|\$\{.+?\})[^>]*>/) && hasExport) {
 				// if (this.getModuleInfo(id).hasJSX) {
 				hasHot = true;
 				after += '\n' + PREFRESH;
 				// }
 			}
 
-			if (!hasHot) return null;
+			if ((!hot && !hasHot) || (!hasExport && !code.match(/\bimport\b/))) return null;
```
developit, line 81: `hasHot` can be true in prod, but we will already have injected `null` for the HMR API on line 52, so we should still bail out here.

```suggestion
			if (!hot || !hasExport) return null;
```

src/plugins/wmr/plugin.js
```diff
@@ -65,33 +66,41 @@ export default function wmrPlugin({ hot = true } = {}) {
...
-				s.append(after);
-				s.prepend(
-					`import { createHotContext as $createHotContext$ } from 'wmr';const $IMPORT_META_HOT$ = $createHotContext$(import.meta.url);${before}`
-				);
+				if (!hasHot) {
+					s.append(`\nimport { createHotContext as $w_h$ } from 'wmr'; $w_h$(import.meta.url);`);
+				} else {
+					s.append(after);
+					s.prepend(
+						`import { createHotContext as $createHotContext$ } from 'wmr';const $IMPORT_META_HOT$ = $createHotContext$(import.meta.url);${before}`
```
developit, line 96: might as well make these the same function as above:

```suggestion
						`import { createHotContext as $w_h $ } from 'wmr';const $IMPORT_META_HOT$ = $w_h $(import.meta.url);${before}`
```

package.json
developit: it came back lol

src/plugins/wmr/plugin.js
```diff
@@ -65,33 +67,42 @@ export default function wmrPlugin({ hot = true } = {}) {
...
-			if (code.match(/html`[^`]*<([a-zA-Z][a-zA-Z0-9.:-]*|\$\{.+?\})[^>]*>/) && code.match(/\bexport\b/)) {
+			if (code.match(/html`[^`]*<([a-zA-Z][a-zA-Z0-9.:-]*|\$\{.+?\})[^>]*>/) && hasExport) {
 				// if (this.getModuleInfo(id).hasJSX) {
 				hasHot = true;
 				after += '\n' + PREFRESH;
 				// }
 			}
 
-			if (!hasHot) return null;
+			if (!hot || !hasEsmKeywords) return null;
```
developit, line 83: third time's a charm?

```suggestion
			if (!hasHot && (!hot || !hasEsmKeywords)) return null;
```

marvinhagemeister (approved):

PR #222, opened 2020-12-04
make small poc for bubbling and keeping track of a module-graph

JoviDeCroock:
Current unknown mainly is whether or not when we are bubbling up if we'll also get updated child-modules.

Parent (with accept callback)
Child (whithout) 

When the child gets an update, we bubble up to Parent. When Parent handles this will the browser have a new version of Child or will we have to rewrite the import in Parent (recursively)?

Current tests seem to indicate that it works

### DIFF
--- a/src/plugins/wmr/plugin.js
+++ b/src/plugins/wmr/plugin.js
@@ -39,7 +39,7 @@ export function getWmrClient({ hot = true } = {}) {
  * @param {object} options
  * @returns {import('rollup').Plugin}
  */
-export default function wmrPlugin({ hot = true } = {}) {
+export default function wmrPlugin({ hot = true, cwd } = {}) {
 	if (BYPASS_HMR) hot = false;
 
 	return {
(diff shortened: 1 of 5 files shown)


src/plugins/wmr/plugin.js
```diff
@@ -90,12 +91,25 @@ export default function wmrPlugin({ hot = true } = {}) {
 			});
 
 			if (hot) {
+				const [imports] = await parse(code, id);
```
developit, line 94: I think we can do this in the middleware instead, since it's already parsing all of the imports?

https://github.com/preactjs/wmr/blob/3fbed71240cd1ae252c2af68192b9b0fb0cad4aa/src/wmr-middleware.js#L286

Something like:

```diff
+		const imports = new Set();

		code = await transformImports(code, id, {
			// ...
			async resolveId(spec, importer) {
				// ...
+				imports.add(spec);
				return spec;
			}
		});
+
+		code += `$IMPORT_META_HOT$.dependencies = ${JSON.stringify(Array.from(imports))}`;
```

This also is nice because it ensures they're all normalized.

JoviDeCroock:
I don't think this is feasible, we can't communicate to the server which chain of modules has to be invalidated when we've bubbled up to a new boundary.

PR #223, opened 2020-12-04
feat: Clean output prior to build

rschristian:
I've seen a few comments that the output directory fills up with old files and experienced this myself when going to quickly deploy something to prod. A manual `rm -rf && wmr build` isn't difficult, but it'd be nice if the tool did this itself.


developit:
Nice - good call.

I just had an idea while pondering this: what if we deleted files from the output directory after building if they weren't written by the current build? Something like:

```js
// src/plugins/cleanup-plugin.js
import { promises as fs } from 'fs';
import { totalist } from 'totalist';

export default function cleanupPlugin() {
	return {
		name: 'cleanup',
		async writeBundle(options, bundle) {
			await totalist(options.dir, (rel, abs) => {
				if (rel in bundle) return;
				// can probably avoid even waiting on these promises
				fs.unlink(abs).catch(String);
			});
		}
	});
}
```

rschristian:
Looks good, yeah. 

Though it would leave empty directories I believe. I'm sure there's some way to nicely clear them, I just need to go looking for it.

developit:
hah - yeah I was pondering the same thing! On the plus side, empty directories are pretty much universally ignored by deployment tooling, but it might look odd. We don't currently generate hashed directories either, so it should be a fairly bounded set.

developit (approved):
I might still try to throw a test together for this.

lukeed:
Just to echo a comment from Slack:

This approach would work find with what wmr offers ootb because its plugins all make use of `this.emitFile` directly. But given that production builds can bring in _any_ Rollup plugin, there's no knowing that the dev's plugins _arent_ using `fs.writeFile`. We've run into this a few times with different plugins in Svelte support.

When this happens, the output file isn't a part of the `bundle` manifest. So any new files from these plugins will be instantly deleted.

rschristian:
>  On the plus side, empty directories are pretty much universally ignored by deployment tooling, but it might look odd. 

Do they? I still default back to rsync for basic things and pruning empty directories is a flag for it. It's probably less of a problem than JS but probably not appreciated either. 

Not sure if there's a better method than recursively checking every directory in the output either.

src/build.js
```diff
@@ -11,6 +12,8 @@ import { setCwd } from './plugins/npm-plugin/registry.js';
 export default async function build(options = {}) {
 	options.out = options.out || 'dist';
 
+	await premove(options.out, { cwd: options.cwd });
```
developit, line 15: nit: I think this needs to happen after the `normalizeOptions()` call on line 20.

developit:
LGTM aside from the nit about running it after `normalizeOptions()`. 👍

src/build.js
lukeed, line 15: I think `out` is always absolute path because of `nornalize-output` so technically the `cwd` option isn't needed but can't hurt... basically a noop

lukeed (approved):

src/build.js
lukeed: Replacing my comment for this. Didn't scroll down 🤦‍♂️

lukeed: Jk

marvinhagemeister (approved):

Issue #221, opened 2020-12-03
Old and new components in DOM on url change with preact-iso Router


EmilTholin:
There is a very brief flicker in the DOM on url change because of both the old and new route children being rendered at the same time when using the `preact-iso` Router. Here it is magnified with a breakpoint:
![old_route_component_in_dom](https://user-images.githubusercontent.com/11573167/101089685-cc6d4d00-35b5-11eb-9fbb-ca2959fbbc5c.gif)

Is there any way of not having the previous children of the Router being rendered? I don't fully understand the code, but I assume it is this way for the awesome support of lazy components.

developit:
Hiya! Yes - this is something I caught before the release but didn't have time to fix. It's relatively straightforward and I'll try to get it addressed ASAP.

PR #224, opened 2020-12-04
fixed(router): lazy router flicker

Coiven:
fixed this issue #221

### DIFF
--- a/demo/public/lib/loc.js
+++ b/demo/public/lib/loc.js
@@ -67,7 +67,7 @@ export function Router(props) {
 	let a = children.filter(c => c.props.path === path);
 	if (a.length == 0) a = children.filter(c => c.props.default);
 	curChildren.current = a.map((p, i) => cloneElement(p, { path, query }));
-	return curChildren.current.concat(prevChildren.current || []);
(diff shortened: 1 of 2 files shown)


developit:
hiya - this fix works for synchronous routing, but breaks asynchronous routing.

developit (changes requested):
hiya - this fix works for synchronous routing, but breaks asynchronous routing. I don't know if we have a page in the demo app that would trigger the async case, but it's an important one. I'll try to add a test as a way to clarify.

What we need to do to fix this in a way that accounts for async is to hide the incoming route until it commits.

Coiven:
> What we need to do to fix this in a way that accounts for async is to hide the incoming route until it commits.

yeah, I agree that this is a solution. But I mean that Route Component will always render twice when route changes. Is it better to wait for the incoming route to be ready before returning than show/hide.

developit:
@Coiven components can't wait/defer.

danielweck:
I believe this was fixed via https://github.com/preactjs/wmr/pull/337
Can this PR now be closed?

PR #233, opened 2020-12-05
WMR-229 node builtins

dbetteridge:
First pass at adding a helpful error message when attempting to use Node built-ins.

Let me know if any changes need making!

### DIFF
--- a/src/plugins/node-builtins-plugin.js
+++ b/src/plugins/node-builtins-plugin.js
@@ -1,31 +1,35 @@
-const BUILTINS = {
-	fs: `export default {}`,
-	http: `export default {}`,
-	https: `export default {}`,
-	url: `export default {}`,
-	util: `export default {}`,
-	zlib: `export default {}`
-};
+import { builtinModules } from 'module';
 
 /**
- * Stub out Node built-ins.
+ * Return an error on Node built-ins in production.
+ * Return a warning and stub Node built-ins otherwise.
  * @param {object} [options]
  * @returns {import('rollup').Plugin}
  */
-export default function nodeBuiltinsPlugin({} = {}) {
+export default function nodeBuiltinsPlugin({ production } = {}) {
 	return {
 		name: 'node-builtins',
 		resolveId(id) {
-			if (BUILTINS.hasOwnProperty(id)) return id;
+			if (builtinModules.includes(id)) {
+				return id;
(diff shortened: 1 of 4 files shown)


developit (commented):
Looks good! Reading this I am wondering: would it be more helpful to warn during development but error when building for production? A bunch of the plugins accept a `production` boolean, and we could have this one use it to fall back to the `export default {}` shim module during development while emitting `this.warn()`.

dbetteridge:
> Looks good! Reading this I am wondering: would it be more helpful to warn during development but error when building for production? A bunch of the plugins accept a `production` boolean, and we could have this one use it to fall back to the `export default {}` shim module during development while emitting `this.warn()`.

Cheers!
I've tacked in a check for a 'production' boolean (which is now passed by the prodBundler) and return a this.warn for test/dev

src/plugins/node-builtins-plugin.js
```diff
@@ -1,31 +1,34 @@
...
-	https: `export default {}`,
-	url: `export default {}`,
-	util: `export default {}`,
-	zlib: `export default {}`
-};
+import { builtinModules } from 'module';
 
 /**
- * Stub out Node built-ins.
+ * Return an error on Node built-ins.
```
rschristian, line 4: Nit: It doesn't always return an error after your latest changes.

dbetteridge: Valid Nit! Will fix

developit (approved):
Beautiful. This looks great to me.

src/plugins/node-builtins-plugin.js
```diff
@@ -1,31 +1,35 @@
...
 		load(id) {
-			if (BUILTINS.hasOwnProperty(id)) {
+			if (builtinModules.includes(id)) {
+				if (production) {
+					this.error(`Error: ${id} is a Node built-in - WMR does not polyfill these`);
+				}
+				this.warn(`
+				Warning: ${id} is a Node built-in - WMR does not polyfill these. 
+				For development the module has been stubbed.
+				`);
```
developit, line 25:
```suggestion
				this.warn(
					`Warning: ${id} is a Node built-in - WMR does not polyfill these.\n` +
					`For development the module has been stubbed.`
				);
```

PR #234, opened 2020-12-06
feat: Adding support for yarn in create-wmr

rschristian:
Adds support for initializing WMR project with Yarn.

Also now properly catches in the chance the chosen package manager is not installed/accessible.

### DIFF
--- a/packages/create-wmr/src/index.js
+++ b/packages/create-wmr/src/index.js
@@ -7,10 +7,11 @@ import ora from 'ora';
 import kleur from 'kleur';
 import pkgInstall from 'pkg-install';
 const { getPackageManager, install } = pkgInstall;
-const { dim, bold, cyan } = kleur;
+const { dim, bold, cyan, red } = kleur;
 
 sade('create-wmr [dir]', true)
 	.option('--eslint', 'Set up the Preact ESLint configuration (takes a lot longer)', false)
+	.option('--yarn', 'Set up WMR using Yarn instead of NPM as the package manager', false)
 	.describe('Initialize a WMR project')
 	.example('npm init wmr ./some-directory')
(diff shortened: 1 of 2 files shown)


JoviDeCroock (approved):
Nice work!

packages/create-wmr/src/index.js
```diff
@@ -7,10 +7,11 @@ import ora from 'ora';
 import kleur from 'kleur';
 import pkgInstall from 'pkg-install';
 const { getPackageManager, install } = pkgInstall;
-const { dim, bold, cyan } = kleur;
+const { dim, bold, cyan, red } = kleur;
 
 sade('create-wmr [dir]', true)
 	.option('--eslint', 'Set up the Preact ESLint configuration (takes a lot longer)', false)
+	.option('--yarn', 'Set up WMR using Yarn instead of NPM as the package manager', false)
```
developit, line 14: I wonder - if the idea is to have this work with `yarn init` / `yarn create`, we could auto-detect that:

```js
const preferYarn = /yarn/.test(process.env.NPM_EXECPATH);
```

rschristian: Sure, could do that. I did it as a flag just to mirror what I think most other tools use (I know that's how it is on CRA & Preact CLI) but this is definitely cleaner. I'll change this + a note on the ReadMe

lukeed:
Thoughts on `pnpm` support while we're here? :)

rschristian:
I've never used pnpm but certainly can take a look? `pkg-install` seems to only support npm & yarn though so might need a bit more than a couple ternaries to get support.

lukeed:
Ah, nvm then. Not a blocker at all

PR #237, opened 2020-12-06
Throw Error when Node doesn't support workers

JoviDeCroock:


marvinhagemeister (approved):

lukeed:
`worker_threads.Worker` was added in Node 10.5.0 – https://nodejs.org/api/worker_threads.html#worker_threads_class_worker

Our `engines` protect this from ever happening

JoviDeCroock:
@lukeed Apparently not, @hassanbazzi was running into it with Node 12

lukeed:
Something else going on then. Saw Hassan was using 12.14 – checked that and an older version for good measure.

Screen Shot 2020-12-06 at 12 10 04 PM

developit:
It's likely related to eval inside a Module worker, but I know for sure that works on 12.9+

PR #245, opened 2020-12-11
fix: Adding jsxFactory to create-wmr's tsconfig file

rschristian:
Enabling `noImplicitAny` or `strict` in a new WMR app with `.tsx` files will present the following error at every usage of JSX: 

```
TS7026: JSX element implicitly has type 'any' because no interface 'JSX.IntrinsicElements' exists.
```

While this won't hinder the app from being built in any way, it is quite distracting in an IDE. This should stop that error from popping up.

### DIFF
--- a/packages/create-wmr/tpl/tsconfig.json
+++ b/packages/create-wmr/tpl/tsconfig.json
@@ -2,6 +2,7 @@
 	"compileOnSave": false,
 	"compilerOptions": {
 		"jsx": "preserve",
+		"jsxFactory": "preact.h",
 		"allowJs": true,
 		"checkJs": true,
 		// "strict": true,
--- a/packages/create-wmr/tpl/tsconfig.json
+++ b/packages/create-wmr/tpl/tsconfig.json
@@ -3,6 +3,7 @@
 	"compilerOptions": {
 		"jsx": "preserve",
 		"jsxFactory": "preact.h",
+		"jsxFragmentFactory": "preact.Fragment",
 		"allowJs": true,
 		"checkJs": true,
 		// "strict": true,


developit:
I think this makes sense, but I also really want to find a way to make it not required. Going to try a couple things and report back.

developit:
Ok I think I got this to work, but it's gross: we'd need to make an ambient shadow namespace for `preact.JSX`:

```ts
declare namespace JSX {
	interface IntrinsicElements extends preact.JSX.IntrinsicElements {}
	interface IntrinsicAttributes extends preact.JSX.IntrinsicAttributes {}
	interface Element extends preact.JSX.Element {}
	interface ElementClass extends preact.JSX.ElementClass {}
	interface ElementAttributesProperty extends preact.JSX.ElementAttributesProperty {}
	interface ElementChildrenAttribute extends preact.JSX.ElementChildrenAttribute {}
	interface CSSProperties extends preact.JSX.CSSProperties {}
	interface SVGAttributes extends preact.JSX.SVGAttributes {}
	interface PathAttributes extends preact.JSX.PathAttributes {}
	interface TargetedEvent extends preact.JSX.TargetedEvent {}
	interface DOMAttributes<Target extends EventTarget> extends preact.JSX.DOMAttributes<Target> {}
	interface HTMLAttributes<RefType extends EventTarget = EventTarget> extends preact.JSX.HTMLAttributes<RefType> {}
}
```

packages/create-wmr/tpl/tsconfig.json
```diff
@@ -2,6 +2,7 @@
 	"compileOnSave": false,
 	"compilerOptions": {
 		"jsx": "preserve",
+		"jsxFactory": "preact.h",
```
developit, line 5:
```suggestion
		"jsxFactory": "preact.h",
		"jsxFragmentFactory": "preact.Fragment",
```

developit (approved):

PR #247, opened 2020-12-11
Use production HTM in production

developit:


JoviDeCroock (approved):

developit:
Something's likely up here, I would have expected a non-zero filesize reduction for the prod bundles.

marvinhagemeister (approved):

src/bundler.js
```diff
@@ -98,7 +98,7 @@ export async function bundleProd({
 			htmlEntriesPlugin({ cwd, publicDir, publicPath: '/' }),
 			publicPathPlugin({ publicPath: '/' }),
 			aliasesPlugin({ aliases, cwd: root }),
-			htmPlugin(),
+			htmPlugin({ production: true }),
```
cristianbote, line 101: By default this is set to true inside here https://github.com/preactjs/wmr/blob/3fbed71240cd1ae252c2af68192b9b0fb0cad4aa/src/plugins/htm-plugin.js#L13. so you

developit: (heh, yes you are right)

developit:
Fix is at developit/htm#189

cristianbote:
@developit pushed a test as well to keep track of this in the future. 👍

developit:
Aces! Thanks for the test @cristianbote.

PR #248, opened 2020-12-11
feat: Adding types for preact-iso

### LINKED ISSUE #243: Add TS typings for preact-iso package
It is currently without and would be useful for TS users.

I'll try to take a crack at this tomorrow evening should I have the time.

rschristian:
Adds types to `preact-iso` package. Closes #243 .

Let me know if anything looks incorrect. I've tested locally in a WMR app with the `tsconfig` having `strict` enabled and everything seemed to work correctly.

### DIFF
--- a/packages/preact-iso/hydrate.d.ts
+++ b/packages/preact-iso/hydrate.d.ts
@@ -1,0 +1,3 @@
+import { VNode } from 'preact';
+
+export default function hydrate(jsx: VNode, parent?: VNode): void;
--- a/packages/preact-iso/hydrate.d.ts
+++ b/packages/preact-iso/hydrate.d.ts
@@ -1,3 +1,3 @@
-import { VNode } from 'preact';
+import { ComponentChild, VNode } from 'preact';
 
-export default function hydrate(jsx: VNode, parent?: VNode): void;
+export default function hydrate(jsx: ComponentChild, parent?: VNode): void;
--- a/packages/preact-iso/index.d.ts
+++ b/packages/preact-iso/index.d.ts
@@ -1,0 +1,4 @@
+import { VNode } from 'preact';
+import { PrerenderOptions } from './prerender';
+
+export default function prerender(vnode: VNode, options?: PrerenderOptions): { html: string, links: Set<string>};
--- a/packages/preact-iso/lazy.d.ts
+++ b/packages/preact-iso/lazy.d.ts
@@ -1,5 +1,5 @@
-import { VNode } from 'preact';
+import { ComponentChildren, VNode } from 'preact';
 
 export default function lazy<T>(load: () => Promise<{ default: T }>): T;
 
-export function ErrorBoundary(props: { children?: VNode }): VNode;
+export function ErrorBoundary(props: { children?: ComponentChildren }): VNode;
--- a/packages/preact-iso/lazy.d.ts
+++ b/packages/preact-iso/lazy.d.ts
@@ -1,4 +1,4 @@
-import { FunctionalComponent, VNode } from 'preact';
+import { VNode } from 'preact';
 
 export default function lazy<T>(load: () => Promise<{ default: T }>): T;
 
--- a/packages/preact-iso/lazy.d.ts
+++ b/packages/preact-iso/lazy.d.ts
@@ -1,0 +1,5 @@
+import { FunctionalComponent, VNode } from 'preact';
+
+export default function lazy<T>(load: () => Promise<{ default: T }>): T;
+
+export function ErrorBoundary(props: { children?: VNode }): VNode;
--- a/packages/preact-iso/router.d.ts
+++ b/packages/preact-iso/router.d.ts
@@ -12,5 +12,7 @@ interface RoutableProps {
 }
 
 declare module 'preact' {
-	export interface Attributes extends RoutableProps {}
+	namespace JSX {
+		interface IntrinsicAttributes extends RoutableProps {}
+	}
 }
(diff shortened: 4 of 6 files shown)


packages/preact-iso/hydrate.d.ts
```diff
@@ -0,0 +1,3 @@
+import { VNode } from 'preact';
+
+export default function hydrate(jsx: VNode, parent?: VNode): void;
```
marvinhagemeister, line 3: Curious if this should be `ComponentChildren` instead of `VNode` similar to our types in core: https://github.com/preactjs/preact/blob/b1a42a1bc61bf1f648e19a41cdf54d235810714c/src/index.d.ts#L228-L231

packages/preact-iso/lazy.d.ts
```diff
@@ -0,0 +1,5 @@
+import { VNode } from 'preact';
+
+export default function lazy<T>(load: () => Promise<{ default: T }>): T;
+
+export function ErrorBoundary(props: { children?: VNode }): VNode;
```
marvinhagemeister, line 5: I think children needs to be typed as `ComponentChildren` here. Otherwise this wouldn't be valid:

```jsx
<ErrorBoundary>foo</ErrorBoundary>
```

packages/preact-iso/router.d.ts
```diff
@@ -0,0 +1,19 @@
+import { VNode } from 'preact';
+
+export function LocationProvider(props: { children?: VNode }): VNode;
```
marvinhagemeister, line 3: Haven't tested, but maybe we can simplify that to

```tsx
export const LocationProvider: FunctionComponent;
```

At least the types in core seems to hint that `children` are typed by default: https://github.com/preactjs/preact/blob/b1a42a1bc61bf1f648e19a41cdf54d235810714c/src/index.d.ts#L86-L90

packages/preact-iso/router.d.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export function Router(props: { onLoadEnd?: () => void, onLoadStart?: () => void, children?: VNode[] }): VNode;
+
+export const useLoc: () => { url: string, path: string, query: Object, route };
```
marvinhagemeister, line 7: I think the `query` property is not an Object constructor, but a record. Something like `Record<string, string>`

marvinhagemeister (commented):
This is a great start! Left some minor comments, let me know what you think 👍

packages/preact-iso/router.d.ts
rschristian: Ah, cheers. I believe this was lifted straight from the JS. Missed it when checking over my work.

rschristian: Do you happen to know what a better type for `route` might be? I'm not too sure of what it's even used for.

marvinhagemeister: It looks like it's of the type `(url: string) => void`.

packages/preact-iso/lazy.d.ts
```diff
@@ -0,0 +1,5 @@
+import { ComponentChildren, VNode } from 'preact';
+
+export default function lazy<T>(load: () => Promise<{ default: T }>): T;
```
developit, line 3: The `.default` property is optional and only used if present, I think this should do the trick:

```suggestion
export default function lazy<T>(load: () => Promise<{ default: T } | T>): T;
```

packages/preact-iso/router.d.ts
```diff
@@ -0,0 +1,19 @@
+import { FunctionComponent, VNode } from 'preact';
+
+export const LocationProvider: FunctionComponent;
+
+export function Router(props: { onLoadEnd?: () => void, onLoadStart?: () => void, children?: VNode[] }): VNode;
```
developit, line 5: I wonder if this should use ComponentChildren?

```suggestion
export function Router(props: { onLoadEnd?: () => void, onLoadStart?: () => void, children?: ComponentChildren }): VNode;
```

packages/preact-iso/index.d.ts
```diff
@@ -0,0 +1,4 @@
+import { VNode } from 'preact';
+import { PrerenderOptions } from './prerender';
+
+export default function prerender(vnode: VNode, options?: PrerenderOptions): { html: string, links: Set<string>};
```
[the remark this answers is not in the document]
rschristian: Sorry, missed that.

packages/preact-iso/router.d.ts
rschristian: I don't believe so? Otherwise something like a number would be a valid child. https://github.com/preactjs/preact/blob/1a0837fe887e2ee5d5b504d62d1738e71d2c76fc/src/index.d.ts#L46-L54

packages/preact-iso/router.d.ts
rschristian: Thanks! I appreciate it.

packages/preact-iso/router.d.ts
developit: That's actually allowed in Preact :D

rschristian: Seemed to break the router though... Might just need an optional chain in there.

developit: Ah! that is a very good point actually! VNode is best then.

rschristian: That's what I was thinking.

sync:
see https://github.com/preactjs/wmr/issues/284, it look like those types aren't being included

PR #249, opened 2020-12-11
fix: Aliases on npm modules

### LINKED ISSUE #178: Aliases work differently for source code and npm packages
I haven't had time to diagnose this yet, but it seems like aliases are applied to source code by transforming imports for those modules, whereas imports for those modules found in npm modules are left unmodified:

```js
// index.js:
import 'react';
// index.js: (output)
import '/@npm/preact/compat';
```
vs:
```js
// node_modules/foo/index.js:
import 'react';
// node_modules/foo/index.js: (output)
import '/@npm/react';  // resolves to a second copy of preact/compat with this URL
```

jstans:
Fixes #178

### DIFF
--- a/src/lib/npm-middleware.js
+++ b/src/lib/npm-middleware.js
@@ -76,7 +76,7 @@ export default function npmMiddleware({ source = 'npm', aliases, optimize, cwd }
 			if (cached) return sendCachedBundle(req, res, cached);
 
 			// const start = Date.now();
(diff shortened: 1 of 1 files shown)


developit:
Nice!! I might try to throw a quick test together for this based on the case reported in #178, but this looks A+.

developit (approved):

PR #256, opened 2020-12-15
create-wmr should pre-install preact

developit:
Might also make sense to have `preact-iso` auto-import `preact/debug` in development mode?


rschristian:
Debug would be nice, agreed.

PR #261, opened 2020-12-17
Hoist JSX namespace to be global by default

### LINKED ISSUE #258: Consider declaring ambient JSX declarations
See https://github.com/preactjs/wmr/pull/245#issuecomment-745603400 - we could re-declare `JSX` as a global namespace to avoid requiring `jsxFactory` be set in `tsconfig.json` / `jsconfig.json`.

developit:
This essentially does what #245 did but for folks who aren't using `create-wmr`.

Fixes #258.

### DIFF
--- a/src/plugins/node-builtins-plugin.js
+++ b/src/plugins/node-builtins-plugin.js
@@ -21,7 +21,7 @@ export default function nodeBuiltinsPlugin({ production } = {}) {
 				}
 				this.warn(
 					`Warning: ${id} is a Node built-in - WMR does not polyfill these.\n` +
-					`For development the module has been stubbed.`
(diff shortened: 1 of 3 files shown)


rschristian:
Unless I'm messing up the process of manually adding these changes to the installed `wmr` module, this doesn't seem to silence Webstorm's complaints.  Will keep investigating though

developit:
@rschristian I wonder if we could add something in the scaffolded `tsconfig.json` to force WMR's types to be imported. Maybe via `include:[]`?

rschristian:
I ran through a few tsconfig options last night, none of them ended up working. 

Will need to look into it deeper eventually.

Issue #257, opened 2020-12-15
RFC: Hot Module Replacement


JoviDeCroock:
## State of the art

The state of the art with implementations like web-dev-server, snowpack and vite is following the [esm-hmr spec](https://github.com/snowpackjs/esm-hmr).
This means that we have an hmr-client that lives on the client-side of the application, this hmr-client registers itself and modules to the server-side websocket.

When an update comes in the server will check whether or not the updated module is accepting updates, when it does it sends down the new copy with `?mtime=x` to invalidate the cache and lets the `accept` function handle the rest.
When the module doesn't accept updates it'll bubble up to find the closest parent that does accept the update (think about boundaries).

The above is handled on the server so this means that it has the traversed chain in-memory, this means that if a non-updating child bubbles up to a grandparent or further the server can use its smarts and rewrite all the traversed imports with `?mtime=x` so the accepting parent doesn't work with stale equivalents of our updated module.

### Stale-child problem

With the above we've bumped into one problem, think about the following scenario, we have a file called `parent.js` and a `child.js`. The `parent.js` imports `child.js` and uses it as a util function.
We update `child.js` to return a value of 2 rather than 1, the module will be refetched, the cache will be busted due to `?mtime=x` and we reexecute the function, the parent will get a value of 2 from this function now.

Now we update the `parent.js` file, this accepts the update, the module is fetched and the cache invalidated due to `?mtime=x`. This will import from `child.js` which in our cache is still the old module which will return 1 rather than 2 this means that the wrong logic will be used with the updated parent.

Notice that with the above logic we always orphan the updated module.

## Current wmr implementation

Currently in wmr we don't bubble for non-accepting children, we always expect a module to handle its own updates which can be hard for scenario's like Preact-hooks, hooks don't represent UI but the usage of said hook in a component does.

wmr currently handles everything on the client, which imposes several challenges, bubbling up is possible, we can visualize our module graph but updating children by appending `?mtime=x` isn't possible due to us having no knowledge about the path we've traversed. We can't just invalidate all imports since this could get really costly.

We could try and check which module comes back and rewrite them on the client but in my opinion that isn't the cleanest way, I'm mainly thinking about whether or not we can get around the child-orphaning issue we're currently seeing in the esm-hmr implementation.

I've been trying some in-between solutions like sending along the chain of imports to `wmr` which could serve as a temporary fix.

Initial POC: https://github.com/preactjs/wmr/pull/222

43081j:
[...]

PR #262, opened 2020-12-18
Handle hmr on the server

JoviDeCroock:
This PR equalises our behavior with snowpack, vite and web-dev-server. They comply to the [esm-hmr spec](https://github.com/snowpackjs/esm-hmr) as pointed out in https://github.com/preactjs/wmr/issues/257 this approach has issues because of always orphaning updated modules, however both React-fast-refresh and Prefresh account for this limitation internally. We can safely ship this while we  think about how to improve this.

Todo:

- [x] get the bi-directional socket to work
- [x] bubbling updates with non-accepting children

implements https://github.com/preactjs/wmr/issues/257

### DIFF
--- a/.changeset/nine-crabs-taste.md
+++ b/.changeset/nine-crabs-taste.md
@@ -1,0 +1,5 @@
+---
+'wmr': minor
+---
+
+Handle hmr on the development server, this enables bubbling in hmr signals that aren't accepted.
--- a/packages/wmr/demo/public/index.tsx
+++ b/packages/wmr/demo/public/index.tsx
@@ -6,7 +6,6 @@ import Home from './pages/home.js';
 import NotFound from './pages/_404.js';
 import Header from './header.tsx';
 // import './style.css';
-import { useCounter } from './useCustomHook.js'
 
 const About = lazy(() => import('./pages/about/index.js'));
 const CompatPage = lazy(() => import('./pages/compat.js'));
@@ -15,12 +14,9 @@ const Files = lazy(() => import('./pages/files/index.js'));
 const Environment = lazy(async () => (await import('./pages/environment/index.js')).Environment);
 
 export function App() {
-	const [count, increment] = useCounter();
 	return (
 		<Loc>
 			<div class="app">
-				{count}
-				<button onClick={increment}>Click me</button>
 				<Header />
 				<ErrorBoundary>
 					<Router>
--- a/packages/wmr/demo/public/index.tsx
+++ b/packages/wmr/demo/public/index.tsx
@@ -6,6 +6,7 @@ import Home from './pages/home.js';
 import NotFound from './pages/_404.js';
 import Header from './header.tsx';
 // import './style.css';
+import { useCounter } from './useCustomHook.js'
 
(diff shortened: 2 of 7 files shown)


src/wmr-middleware.js
```diff
@@ -119,6 +121,7 @@ export default function wmrMiddleware({
 		pendingChanges.clear();
 	}
 	watcher.on('change', filename => {
+		console.log('changed', filename);
```
JoviDeCroock, line 124: When we get a change we'll have to look at our module-graph and see if the file accepts updates:

```
function bubbleUpdates() {
  const mod = moduleGraph.get(filename);
  if (mod.acceptsUpdates) {
    // delete build cache;
    return sendUpdateSignal()
  } else if (mod.dependents.size) {
    // iterate over dependents and call bubbleUpdates && delete build cache &&
    // set needsUpdate: true on the module in the moduleGraph
    return;
  } else {
    return sendReloadSignal();
  }
  return sendReloadSignal();
}
```

When we get a request for a file that contains `?t=x` we know that we need to check the imports, we traverse the imports with our usual `resolveId`, if we see a module from the moduleGraph that has a truthy value for needsUpdate we need to append `?t=x` to that import to invalidate that chain of imports.

src/lib/websocket-server.js
```diff
@@ -9,9 +10,41 @@ export default class WebSocketServer extends ws.Server {
...
+	connectClient(client) {
+		this.hmrClients.add(client);
+	}
+
+	disconnectClient(client) {
+		this.hmrClients.delete(client);
+	}
+
+	registerListener(client) {
+		client.on('message', function (data) {
```
JoviDeCroock, line 36: This never fires for some reason

src/wmr-middleware.js
```diff
@@ -339,6 +374,19 @@ export const TRANSFORMS = {
 					spec = `/@npm/${meta.module}${meta.path ? '/' + meta.path : ''}`;
 				}
 
+				const modSpec = spec.replace('../', '/').replace('./', '/');
```
JoviDeCroock, line 377: I don't think this is a good solution, if anyone knows a better one feel free to let me know

src/wmr-middleware.js
```diff
@@ -118,16 +120,44 @@ export default function wmrMiddleware({
...
 	}
+
+	function bubbleUpdates(filename) {
+		// Delete file from the in-memory cache:
+		WRITE_CACHE.delete(filename);
+
+		filename = '/' + filename;
+		const mod = moduleGraph.get(filename);
+
+		if (!mod) return false;
```
JoviDeCroock, line 131: This shouldn't occur but better be safe than sorry 😅

src/plugins/wmr/client.js
```diff
@@ -114,11 +124,19 @@ function getMod(url) {
 	return mod;
 }
 
+let queue = [];
```
ForsakenHarmony, line 127: I think this should be defined in line 9, so it gets defined before use, even though that might not matter here

marvinhagemeister (approved):

PR #270, opened 2020-12-20
create-wmr: Limiting template's tsconfig's scope to "public/"

rschristian:
When there's no `include` flag for the TSConfig it will default to covering all files with supported extensions. As `allowJS` is enabled, this includes JS.

The issue with this comes from enabling `strict`/`noImplicitAny`. Top level config files (like `wmr.config.js`) will now start to throw errors because of missing types. While you can write JSDoc to silence this, it is a bit unnecessary.

The better default would likely be to limit the TSConfig's scope to just the `public/` dir. Top-level configs or other files should likely be excluded.

### DIFF
--- a/packages/create-wmr/tpl/tsconfig.json
+++ b/packages/create-wmr/tpl/tsconfig.json
@@ -15,6 +15,7 @@
 		"allowSyntheticDefaultImports": true,
 		"downlevelIteration": true
 	},
+	"include": ["public"],
 	"typeAcquisition": {
 		"enable": true
 	}


developit:
I'm split on this. I actually prefer all JS files to be validated via checkJS, but it's (very) possible I'm in the minority there. Plus, since we don't scaffold `wmr.config.js` that means folks have to somehow figure out the JSDoc incantation on their own, which is a bit unlikely.

Another option here (not necessarily advocating for it, just pondering) would be to add `wmr.config.js` to `exclude`?

rschristian:
Certainly, and as it may be more opinionated than I first thought, it's simple enough to leave it to the end user. Single field that's largely self-explanatory after all. 

Just anecdotally, most projects I've seen I believe exclude, especially if they do want to use strict mode. JSDoc goes out of date too quickly and there's a whole bunch of config files that just gain nothing from being included, except now the (already fickle) TS lang server has to track and manage.

Though to even have to have this issue the user would need to be in the config anyways to enable `strict`/`noImplicitAny`. Don't know! Just some thoughts.

rschristian:
Closing as this is trivial and a matter of opinion. Doesn't need to stay open.

PR #274, opened 2020-12-21
preact-iso: Adding "useLocation" alias for "useLoc"

### LINKED ISSUE #269: [preact-iso] Why `useLoc` instead of `useLocation`?
Official hooks generally aren't named like this. Is it temporary or was there some issue with using `useLocation`?

Great project by the way, it was super quick to set everything up, right now I'm waiting for the preact-iso type definitions to be released.

rschristian:
Addresses & closes #269 

Adds a simple alias `useLocation()` for `useLoc()`. React-router, Reach, and Wouter all use `useLocation()` so I imagine this is more comfortable for a pretty large group of people.

### DIFF
--- a/packages/preact-iso/router.d.ts
+++ b/packages/preact-iso/router.d.ts
@@ -4,7 +4,7 @@ export const LocationProvider: FunctionComponent;
 
 export function Router(props: { onLoadEnd?: () => void, onLoadStart?: () => void, children?: VNode[] }): VNode;
 
-type LocationHook =	() => { url: string, path: string, query: Record<string, string>, route: (url: string) => void };
+type LocationHook = () => { url: string, path: string, query: Record<string, string>, route: (url: string) => void };
 export const useLoc: LocationHook;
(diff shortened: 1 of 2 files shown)


packages/preact-iso/router.d.ts
```diff
@@ -4,7 +4,9 @@ export const LocationProvider: FunctionComponent;
 
 export function Router(props: { onLoadEnd?: () => void, onLoadStart?: () => void, children?: VNode[] }): VNode;
 
-export const useLoc: () => { url: string, path: string, query: Record<string, string>, route: (url: string) => void };
+type LocationHook = () => { url: string, path: string, query: Record<string, string>, route: (url: string) => void };
+export const useLoc: LocationHook;
+export const useLocation: LocationHook;
```
rschristian, line 9: This could also simply be `... useLocation: typeof useLoc;` and we could get rid of the extra type definiton. I have no real opinions on which is nicer.

JoviDeCroock (approved):

Issue #246, opened 2020-12-11
HMR doesn't work on Render's root component


mattlockyer:
Changing anything inside App component, does not "hot reload".

e.g. Hello Foo -> Hello World / Test -> Test Bar (changed reflected only after manually refreshing page)
```
function App() {
    return <>
        <h1>Hello Foo!</h1>
        <div>
            <p>
                Test
            </p>
        </div>
        <Test />
    </>;
}
```

Changing contents of Test component (code not shown) works fine, hot reloaded.

developit:
Hiya - this will get fixed once we integrate the Babel portion of [prefresh](https://github.com/jovidecroock/prefresh). For now though, you can use `preact-iso` to make it work by default:

```diff
-import { render } from 'preact';
+import { hydrate } from 'preact-iso';

function App() {
    return <>
        <h1>Hello Foo!</h1>
        <div>
            <p>
                Test
            </p>
        </div>
        <Test />
    </>;
}

hydrate(<App />);
```

mattlockyer:
Odd... adding export to main component makes it work... Bizarre!
```
export function App() {
	return <h1>Hello World</h1>;
}
```

sortofsleepy:
Apologies for the possibly silly question -
I'm running into a similar problem here; but with child components. Like Matt noted, I had to add `export` to the component being put into `render`/`hydrate`  but I am finding that changes to child components don't appear to register. Take for instance

```javascript
....
export function App(){
    return <main>
        <Home/>
    </main>
}
hydrate(<App/>)
```

If I change the `Home` component, the change is not noted or written; even if I refresh, the last thing from the cache gets loaded. 

Any ideas? 

Not sure if it matters but in case it helps, running on Windows without using WSL(which has it's own issues apparently but not a requirement for my project)

developit:
@sortofsleepy that seems like a different issue, perhaps relating to watching files on Windows. Do you see any logs in the console at all when saving the file?

sortofsleepy:
> 
> 
> @sortofsleepy that seems like a different issue, perhaps relating to watching files on Windows. Do you see any logs in the console at all when saving the file?

I was wondering if that was the case at first but thought it was weird that I was still seeing it after going back to the native Windows command prompt. In any case no errors appear or anything like that as far as I can tell; things just don't refresh.

davidmz:
I have the same issue as @sortofsleepy (also on Windows). No messages appear in the console. How can I turn on some debug output?

[...]

PR #278, opened 2020-12-21
normalize paths for windows nested hmr

JoviDeCroock:
Fixes https://github.com/preactjs/wmr/issues/246


developit (approved):

developit:
LGTM. I wasn't super sure whether we should be normalizing these, but from what you mentioned the other day they're being compared against unix paths elsewhere. We can always revisit that later.

PR #281, opened 2020-12-21
[preact-iso] Make lazy() thrown promise resolve to component

### LINKED ISSUE #280: Modify lazy to return component from promise
It would be great to modify https://github.com/preactjs/wmr/blob/master/packages/preact-iso/lazy.js#L7 to return the component as this exposes some great SSR opportunities like collecting chunks for preloading as part of the suspense cycle instead of a HOC / context: https://github.com/jacob-ebey/preact-webpack-prerender/blob/main/webpack/prerender.js#L30

Would this have any adverse effects on performance? I believe this is a first for this method of preload chunk collection. This would also make cross-container module-federation SSR and preloading possible.

I've POC'd this at: https://github.com/jacob-ebey/preact-webpack-prerender/blob/main/webpack/lazy.js#L50

developit:
Fixes #280.

### DIFF
--- a/packages/preact-iso/lazy.js
+++ b/packages/preact-iso/lazy.js
@@ -4,10 +4,10 @@ import { useState, useRef } from 'preact/hooks';
 export default function lazy(load) {
 	let p, c;
 	return props => {
-		if (!p) p = load().then(m => ((c = (m && m.default) || m), 1));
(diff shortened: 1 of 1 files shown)


jacob-ebey (approved):

jacob-ebey:
🚢 it <3 Thank you!

JoviDeCroock (approved):

PR #290, opened 2020-12-31
refactor: Insert user-defined plugins before core plugins

### LINKED ISSUE #289: Using absolute imports within project
As the title suggests, I'm wondering if it's possible to use absolute imports given the nature of the automatically installed NPM modules.

Just for clarity, given the following directory structure, I'd like `index.js` to be able to import `header.js` via `import ... from 'components/header';`. Absolute imports with "public" as a base directory being assumed.

```
public/
--components/
----header.js
--index.js
```

I tried out a few Rollup plugins that were supposed to add that functionality (`rollup-plugin-includepaths` and `@rollup/plugin-node-resolve`) as well as rebuilding from source with an edit to the existing `node-resolve` plugin though I couldn't get any of those to work. Certainly possible it's user error, though it seems WMR will fall back to installing rather than use the plugins. Not entirely sure why.

I can only speak for myself but I know this is something I immediately add in every tool. I imagine some other users would look for this too, so probably a worthwhile addition to [the recipes](https://github.com/preactjs/wmr#recipes) if it is something that's possible currently. I'll certainly keep playing around with this but I thought I'd throw in this issue now.

rschristian:
## Summary

Closes #289 (in a roundabout way)

This allows for using a prefix for absolute paths. Normally this would just require a plugin, but due to the way plugins were inserted, this was not possible. Therefore this PR ensures that user plugins come before WMR core plugins.

A config recipe for #289 can be added to the wiki after this is merged/finalized.

## Details

Concats WMR's plugins on top of user-defined plugins, rather than the other way around. 

A test was added for the absolute import config. Not entirely sure if it's necessary or not; I can see value in testing the config recipes that make up the wiki (Preact CLI has had a couple out-dated ones from time-to-time, resulting in confused users) but at the same time this isn't core functionality.

### DIFF
--- a/src/bundler.js
+++ b/src/bundler.js
@@ -41,7 +41,7 @@ const pathToPosix = p => p.split(sep).join(posix.sep);
  * @property {Record<string, string>} [aliases] module aliases
  * @property {boolean} [profile] Enable bundler performance profiling
  * @property {Record<string, string>} [env]
- * @property {import('rollup').Plugin[]} [plugins]
+ * @property {import('rollup').Plugin[]} plugins
  * @property {Output | Output[]} [output]
  * @property {(error: BuildError)=>void} [onError]
  * @property {(error: BuildEvent)=>void} [onBuild]
@@ -91,7 +91,7 @@ export async function bundleProd({
 		perf: !!profile,
 		preserveEntrySignatures: 'allow-extension',
 		manualChunks: npmChunks ? extractNpmChunks : undefined,
-		plugins: [
+		plugins: plugins.concat([
 			nodeBuiltinsPlugin({ production: true }),
 			externalUrlsPlugin(),
 			sucrasePlugin({
@@ -125,7 +125,7 @@ export async function bundleProd({
 			optimizeGraphPlugin({ publicPath }),
 			minify && minifyCssPlugin({ sourcemap }),
 			copyAssetsPlugin({ cwd })
-		].concat(plugins || [])
+		])
 	});
 
 	/** @type {import('rollup').OutputOptions} */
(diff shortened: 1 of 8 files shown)


developit:
Were you getting errors before making the regex changes? I would have thought both of those regexes should intentionally not match so that the npm and disk plugins don't try to process `resolveId('~/foo')`.

rschristian:
Yes, `Invalid specifier ~/x` and upon fixing that it went searching for an NPM package `~` due to the disk regex.

Is there a better way to go about this?

developit:
I think we can try inserting user-defined plugins before the core ones. That feels more intuitive anyway.

rschristian:
Sorry for the delay, I'll try to take a look at this later, after work.

src/bundler.js
```diff
@@ -41,7 +41,7 @@ const pathToPosix = p => p.split(sep).join(posix.sep);
  * @property {Record<string, string>} [aliases] module aliases
  * @property {boolean} [profile] Enable bundler performance profiling
  * @property {Record<string, string>} [env]
- * @property {import('rollup').Plugin[]} [plugins]
+ * @property {import('rollup').Plugin[]} plugins
```
rschristian, line 44: This I'm less than sure about, maybe some case I'm not seeing?

Right now [`normalize-options` ensures `plugins` is never undefined](https://github.com/preactjs/wmr/blob/master/src/lib/normalize-options.js#L17), so as far as I can tell there's no reason for it to be optional. Not a big deal to have a empty array to fallback to (as was the case before this PR) but I don't see a need for it currently.

JoviDeCroock (approved):

JoviDeCroock:
@rschristian would you mind rebasing this one, sorry for the delay

PR #296, opened 2021-01-03
Tighten up ESM_KEYWORDS

### LINKED ISSUE #295: Non module script incorrectly detected as module
Looks like the [ESM_KEYWORD](https://github.com/preactjs/wmr/blob/2024c9bd8e1ffa7192a37ecea7de53db3695f7e9/src/plugins/fast-cjs-plugin.js#L3) regex isn't specific enough. It matches the following code and thinks that this is a module file:

```js
(function () {
	'use strict';
	// import statements are fun <- this should not be detected as a module file
	window.out.textContent = 'should work';
})();
```

Another snippet that causes the same issue:

```jsx
function foo() {
  // The next line is not an indicator that this is a module file
  return "<div>Add the import to your file</div>"
}
```

The incorrect detection leads to the following error:

```txt
SyntaxError: Cannot use import statement outside a module
```

Workaround: Downgrading to `wmr@0.0.11` works.

developit:
Fixes #295.

### DIFF
--- a/src/plugins/fast-cjs-plugin.js
+++ b/src/plugins/fast-cjs-plugin.js
@@ -1,6 +1,6 @@
 const CJS_KEYWORDS = /\b(module\.exports|exports)\b/;
 
-export const ESM_KEYWORDS = /(\bimport\s*(\{.*?\}\s*from|\s[\w$]\s+from)?\s*['"]|[\s;]export(\s+(default|const|var|let)[^\w$]|\s*\{))/;
+export const ESM_KEYWORDS = /(\bimport\s*(\{.*?\}\s*from|\s[\w$]+\s+from)?\s*['"]|[\s;]export(\s+(default|const|var|let|function|class)[^\w$]|\s*\{))/;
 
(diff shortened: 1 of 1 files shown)


src/plugins/fast-cjs-plugin.js
```diff
@@ -1,6 +1,6 @@
 const CJS_KEYWORDS = /\b(module\.exports|exports)\b/;
 
-export const ESM_KEYWORDS = /(\bimport\s*(\{|\s['"\w_$])|[\s;]export(\s+(default|const|var|let)[^\w$]|\s*\{))/;
+export const ESM_KEYWORDS = /(\bimport\s*(\{.*?\}\s*from|\s[\w$]\s+from)?\s*['"]|[\s;]export(\s+(default|const|var|let)[^\w$]|\s*\{))/;
```
marvinhagemeister, line 3: This doesn't match a default export like:

```js
import foo from "bar";
```

I think your suggestion on slack to experiment with alternative solutions is better for the long term. A regex can always lead to false positives :+1:

developit: I didn't suggesting something other than a regex :P

JoviDeCroock (approved):

Issue #298, opened 2021-01-05
Using parameters with `preact-iso/router`


smeijer:
When using `preact-iso/router`, which is the default in `wmr`, it doesn't seem to be possible to use route parameters.

Am I missing something? What's the recommended approach here?

```jsx
import hydrate from 'preact-iso/hydrate';
import { LocationProvider, Router } from 'preact-iso/router';
import Home from './pages/home/index.js';
import NotFound from './pages/_404.js';

export function App() {
  return (
    <LocationProvider>
      <div class="app">
        <ErrorBoundary>
          <Router>
            <Home path="/:page?" />
            <NotFound default />
          </Router>
        </ErrorBoundary>
      </div>
    </LocationProvider>
  );
}
```

The sample will always result in "not found". Removing the optional parameter from the home route, will make "home" work again.

```diff
-            <Home path="/:page?" />
+            <Home path="/" />
```

JoviDeCroock:
Hey @smeijer 

This has been implemented in https://github.com/preactjs/wmr/pull/300 not merged/released yet but if you want you can try it out

smeijer:
Thanks for the heads up. I'll give it a shot.

orzechowskid:
thanks for this issue and for the PR!  I noticed that the following code functions correctly on `main` but on the `wildcard-routes` branch will throw a `TypeError` and claim that vnode.props is undefined:

```javascript
const publicRoutes = {
  "/blog": lazy(() => import('./pages/blog.js')),
  "/blog/:id": lazy(() => import('./pages/blog/single-post.js'))
};
const privateRoutes = {
  "/settings": lazy(() => import('./pages/settings.js'))
};
export function App(serverData) {
  return (
        <main id="main">
          <Router>
            {Object.entries(publicRoutes).map(
              ([ path, Component ]) => <Component path={path} key={path} />
            )}
            {Object.entries(privateRoutes).map(
              ([ path, Component ]) => <PrivateRoute component={Component} path={path} key={path} />
            )}
            <NotFound default />
          </Router>
        </main>
  );
}
```
I believe this is due to `<Router>`'s child array not being flattened before `exec()` is called.  the workaround is trivial of course but I figured I'd mention it anyway.

thanks again!

PR #300, opened 2021-01-06
(preact-iso) - support route parameters

JoviDeCroock:
Resolves: https://github.com/preactjs/wmr/issues/298

### DIFF
--- a/packages/preact-iso/README.md
+++ b/packages/preact-iso/README.md
@@ -100,7 +100,7 @@ const App = () => (
 );
 ```
 
-During prerendering, the generated HTML includes our full `<Home>` and `<Profile>` component output because it waits for the `lazy()`-wrapped `import()` to resolve.
+During prerendering, the generated HTML includes our full `<Home>` and `<Profiles>` component output because it waits for the `lazy()`-wrapped `import()` to resolve.
 
 **Progressive Hydration:** When the app is hydrated on the client, the route (`Home` or `Profile` in this case) suspends. This causes hydration for that part of the page to be deferred until the route's `import()` is resolved, at which point that part of the page automatically finishes hydrating.
 
--- a/packages/preact-iso/README.md
+++ b/packages/preact-iso/README.md
@@ -84,7 +84,7 @@ import { LocationProvider, Router, useLoc } from 'preact-iso/router';
 
 // Asynchronous (throws a promise)
 const Home = lazy(() => import('./routes/home.js'));
-const Profiles = lazy(() => import('./routes/profile.js'));
+const Profile = lazy(() => import('./routes/profile.js'));
 const Profiles = lazy(() => import('./routes/profiles.js'));
 
 const App = () => (
--- a/packages/preact-iso/README.md
+++ b/packages/preact-iso/README.md
@@ -66,7 +66,9 @@ const { html, links } = await prerender(<App />, { maxDepth: 10 });
 import hydrate from 'preact-iso/hydrate';
 
 const App = () => (
-	<div class="app"><h1>Hello World</h1></div>
+	<div class="app">
+		<h1>Hello World</h1>
+	</div>
 );
 
 hydrate(<App />);
@@ -82,14 +84,16 @@ import { LocationProvider, Router, useLoc } from 'preact-iso/router';
 
 // Asynchronous (throws a promise)
 const Home = lazy(() => import('./routes/home.js'));
-const Profile = lazy(() => import('./routes/profile.js'));
+const Profiles = lazy(() => import('./routes/profile.js'));
+const Profiles = lazy(() => import('./routes/profiles.js'));
 
 const App = () => (
 	<LocationProvider>
 		<ErrorBoundary>
 			<Router>
 				<Home path="/" />
-				<Profile path="/profile" />
+				<Profiles path="/profiles" />
+				<Profile path="/profiles/:id" />
 			</Router>
 		</ErrorBoundary>
 	</LocationProvider>
--- a/packages/preact-iso/router.js
+++ b/packages/preact-iso/router.js
@@ -111,7 +111,7 @@ export function Router(props) {
 
 	if (curChildren.current.length === 0) curChildren.current = props.children.filter(x => x.props.default)
 
-	return curChildren.current.concat(prevChildren.current || []);
+	return [...curChildren.current[0], ...(prevChildren.current || [])];
 }
 
 Router.Provider = LocationProvider;
--- a/packages/preact-iso/router.js
+++ b/packages/preact-iso/router.js
@@ -1,6 +1,8 @@
 import { h, createContext, cloneElement } from 'preact';
 import { useContext, useMemo, useReducer, useEffect, useRef } from 'preact/hooks';
 
+const EMPTY = {};
+
 const UPDATE = (state, url, push) => {
 	if (url && url.type === 'click') {
 		const link = url.target.closest('a[href]');
@@ -17,6 +19,70 @@ const UPDATE = (state, url, push) => {
 	else if (push === false) history.replaceState(null, '', url);
 	return url;
 };
+
+const exec = (url, route, opts) => {
+	let reg = /(?:\?([^#]*))?(#.*)?$/,
+		c = url.match(reg),
+		matches = {},
+		ret;
+
+	if (c && c[1]) {
+		let p = c[1].split('&');
+		for (let i=0; i<p.length; i++) {
+			let r = p[i].split('=');
+			matches[decodeURIComponent(r[0])] = decodeURIComponent(r.slice(1).join('='));
+		}
+	}
+
+	url = segmentize(url.replace(reg, ''));
+	route = segmentize(route || '');
+	let max = Math.max(url.length, route.length);
+	for (let i=0; i<max; i++) {
+		if (route[i] && route[i].charAt(0)===':') {
+			let param = route[i].replace(/(^:|[+*?]+$)/g, ''),
+				flags = (route[i].match(/[+*?]+$/) || EMPTY)[0] || '',
+				plus = ~flags.indexOf('+'),
+				star = ~flags.indexOf('*'),
+				val = url[i] || '';
+
+			if (!val && !star && (flags.indexOf('?')<0 || plus)) {
+				ret = false;
+				break;
+			}
+
+			matches[param] = decodeURIComponent(val);
+			if (plus || star) {
+				matches[param] = url.slice(i).map(decodeURIComponent).join('/');
+				break;
+			}
+		}
+		else if (route[i] !== url[i]) {
+			ret = false;
+			break;
+		}
+	}
+
+	if (opts.default!==true && ret===false) return false;
+
+	return matches;
+}
+
+const segmentize = (url) => url.replace(/(^\/+|\/+$)/g, '').split('/');
+const rankSegment = (segment) => segment.charAt(0)==':' ? (1 + '*+?'.indexOf(segment.charAt(segment.length-1))) || 4 : 5;
+const rank = (path) => segmentize(path).map(rankSegment).join('');
+const rankChild = (vnode) => vnode.props.default ? 0 : rank(vnode.props.path);
+
+const pathRankSort = (a, b) => (
+	(a.rank < b.rank) ? 1 :
+		(a.rank > b.rank) ? -1 :
+			(a.index - b.index)
+);
+
+const prepareVNodeForRanking = (vnode, index) => {
+	vnode.index = index;
+	vnode.rank = rankChild(vnode);
+	return vnode.props;
+}
 
 export function LocationProvider(props) {
 	const [url, route] = useReducer(UPDATE, location.pathname + location.search);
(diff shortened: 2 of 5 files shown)


rschristian:
Might be useful to add a route with parameters to [the `preact-iso` ReadMe](https://github.com/preactjs/wmr/blob/master/packages/preact-iso/README.md#routerjs) too.

packages/preact-iso/router.js
```diff
@@ -82,13 +148,13 @@ export function Router(props) {
 		} else commit();
 	}, [url]);
 
-	const children = [].concat(...props.children);
-
-	let a = children.filter(c => c.props.path === path);
-
-	if (a.length == 0) a = children.filter(c => c.props.default);
+	curChildren.current = props.children
```
developit, line 135: `<Router><div default>a</div></Router>` will produce `props.children={type:'div',children:'a'}` which breaks here.

packages/preact-iso/README.md
```diff
@@ -82,21 +84,23 @@ import { LocationProvider, Router, useLoc } from 'preact-iso/router';
 
 // Asynchronous (throws a promise)
 const Home = lazy(() => import('./routes/home.js'));
-const Profile = lazy(() => import('./routes/profile.js'));
+const Profiles = lazy(() => import('./routes/profile.js'));
```
rschristian, line 87:
```suggestion
const Profile = lazy(() => import('./routes/profile.js'));
```

Typo I believe

packages/preact-iso/router.js
```diff
@@ -18,6 +20,45 @@ const UPDATE = (state, url, push) => {
 	return url;
 };
 
+const segmentize = (url) => url.replace(/(^\/+|\/+$)/g, '').split('/');
+const exec = (url, route, opts) => {
+	let reg = /(?:\?([^#]*))?(#.*)?$/,
+		matches = {},
+		ret;
+
+	url = segmentize(url.replace(reg, ''));
```
developit, line 29:
```suggestion
	let matches = {},
		ret;

	url = segmentize(url);
```

packages/preact-iso/router.js
```diff
@@ -82,13 +122,11 @@ export function Router(props) {
 		} else commit();
 	}, [url]);
 
-	const children = [].concat(...props.children);
-
-	let a = children.filter(c => c.props.path === path);
-
-	if (a.length == 0) a = children.filter(c => c.props.default);
+	curChildren.current = props.children
+		.map(vnode => exec(url, vnode.props.path, vnode.props) ? cloneElement(vnode, { path, query }) : null)
```
developit, line 126:
```suggestion
		.map(vnode => exec(value.path, vnode.props.path, vnode.props) ? cloneElement(vnode, { path, query }) : null)
```

packages/preact-iso/router.js
```diff
@@ -18,6 +20,44 @@ const UPDATE = (state, url, push) => {
 	return url;
 };
 
+const segmentize = (url) => url.replace(/(^\/+|\/+$)/g, '').split('/');
+const exec = (url, route, opts) => {
+	let matches = {},
+		ret;
+
+	url = segmentize(url);
+	route = segmentize(route || '');
```
developit, line 29: `url` is already normalized now, so I think this can be shrunk to:

```suggestion
const exec = (url, route, opts) => {
	let matches = {},
		ret;

	url = url.slice(1).split('/');
	route = (route || '').slice(1).replace(/\/$/, '').split('/');
```

packages/preact-iso/router.js
```diff
@@ -82,13 +121,11 @@ export function Router(props) {
 		} else commit();
 	}, [url]);
 
-	const children = [].concat(...props.children);
-
-	let a = children.filter(c => c.props.path === path);
-
-	if (a.length == 0) a = children.filter(c => c.props.default);
+	curChildren.current = props.children
+		.map(vnode => exec(path, vnode.props.path, vnode.props) ? cloneElement(vnode, { path, query }) : null)
```
developit, line 125:
```suggestion
		.map(vnode => exec(path, vnode.props.path, vnode.props) && cloneElement(vnode, { path, query }))
```

packages/preact-iso/router.js
```diff
@@ -18,6 +20,43 @@ const UPDATE = (state, url, push) => {
...
+	let matches = {},
+		ret;
+
+	url = url.trim('/').split('/');
+	route = (route || '').trim('/').split('/');
+	let max = Math.max(url.length, route.length);
+	for (let i=0; i<max; i++) {
+		if (route[i] && route[i].charAt(0)===':') {
+			let param = route[i].replace(/(^:|[+*?]+$)/g, ''),
+				flags = (route[i].match(/[+*?]+$/) || EMPTY)[0] || '',
```
developit, line 33:
```suggestion
				flags = route[i].match(/[+*?]*$/)[0],
```

developit, line 33:
```suggestion
		if (route[i] && route[i][0]==':') {
			let flags = route[i].match(/[+*?]*$/)[0],
				param = route[i].slice(1, -flags.length),
```

developit: or:

```suggestion
		const m = route[i] && route[i].match(/^:(.+?)([+*?]*)$/);
		if (m) {
			let [, param, flags] = m,
```

packages/preact-iso/router.js
```diff
@@ -18,6 +20,43 @@ const UPDATE = (state, url, push) => {
...
+			}
+		}
+		else if (route[i] !== url[i]) {
+			ret = false;
+			break;
+		}
+	}
+
+	if (opts.default!==true && ret===false) return false;
+
```
developit, line 56:
```suggestion
```

packages/preact-iso/router.js
```diff
@@ -18,6 +20,43 @@ const UPDATE = (state, url, push) => {
...
+			matches[param] = decodeURIComponent(val);
+			if (plus || star) {
+				matches[param] = url.slice(i).map(decodeURIComponent).join('/');
+				break;
+			}
+		}
+		else if (route[i] !== url[i]) {
+			ret = false;
+			break;
```
developit, line 51:
```suggestion
			return;
```

packages/preact-iso/router.js
```diff
@@ -18,6 +20,43 @@ const UPDATE = (state, url, push) => {
...
+		if (route[i] && route[i].charAt(0)===':') {
+			let param = route[i].replace(/(^:|[+*?]+$)/g, ''),
+				flags = (route[i].match(/[+*?]+$/) || EMPTY)[0] || '',
+				plus = ~flags.indexOf('+'),
+				star = ~flags.indexOf('*'),
+				val = url[i] || '';
+
+			if (!val && !star && (flags.indexOf('?')<0 || plus)) {
+				ret = false;
+				break;
```
developit, line 40:
```suggestion
				return;
```

packages/preact-iso/router.js
```diff
@@ -18,6 +20,43 @@ const UPDATE = (state, url, push) => {
 	return url;
 };
 
+const exec = (url, route, opts) => {
```
developit, line 23:
```suggestion
const exec = (url, route) => {
```

packages/preact-iso/router.js
```diff
@@ -82,13 +121,11 @@ export function Router(props) {
 		} else commit();
 	}, [url]);
 
-	const children = [].concat(...props.children);
-
-	let a = children.filter(c => c.props.path === path);
-
-	if (a.length == 0) a = children.filter(c => c.props.default);
+	curChildren.current = props.children
+		.map(vnode => exec(path, vnode.props.path, vnode.props) && cloneElement(vnode, { path, query }))
```
developit, line 125:
```suggestion
		.map((vnode, m) => (m = exec(path, vnode.props.path)) && cloneElement(vnode, { path, query, ...m }))
```

developit (commented):
Here's a few simplifications.

smeijer:
I've been able to try this branch, and the routing part works great! Can't wait to see this being merged.

But... the `useLocation` hook does return these parameters. I think being able to retrieve and use these parameters, is what would make this truly useful.

JoviDeCroock:
@smeijer You mean like `useLocation().parameters` would contain all the `:id`, ....

packages/preact-iso/router.js
```diff
@@ -82,15 +105,13 @@ export function Router(props) {
...
-	if (a.length == 0) a = children.filter(c => c.props.default);
+	curChildren.current = [].concat(props.children || [])
+		.map((vnode, m) => exec(path, vnode.props.path, m = { path, query }) && cloneElement(vnode, m))
+		.filter(Boolean);
 
-	curChildren.current = a.map((p, i) => cloneElement(p, { path, query }));
+	if (curChildren.current.length === 0) curChildren.current = props.children.filter(x => x.props.default)
 
-	return curChildren.current.concat(prevChildren.current || []);
+	return [curChildren.current[0], ...(prevChildren.current || [])];
```
developit, line 114:
```suggestion
	let p, d, m;
	[].concat(props.children || []).some(vnode =>
		exec(path, vnode.props.path, m = { path, query }) ?
			p = cloneElement(vnode, m) :
			(vnode.props.default ? d = cloneElement(vnode, m) : 0, undefined)
	);
	
	return [curChildren.current = p || d, prevChildren.current];
```

smeijer:
> @smeijer You mean like `useLocation().parameters` would contain all the `:id`, ....

Exactly! Extending the example from the docs in this PR:

```js
const App = () => (
	<LocationProvider>
		<ErrorBoundary>
			<Router>
				<Home path="/" />
				<Profiles path="/profiles" />
				<Profile path="/profiles/:id" />
				<Profile path="/profiles/:id/:tab" />
			</Router>
		</ErrorBoundary>
	</LocationProvider>
);
```

When on `example.com/profiles/123`, `useLocation().params` would be

```js
{
  id: '123',
}
```

When on `example.com/profiles/123/hobbies`, `useLocation().params` would be

```js
{
  id: '123',
  tab: 'hobbies',
}
```

That's matching behavior with [`react-router#useParams`](https://reactrouter.com/web/api/Hooks/useparams). Having a separate hook for it named `useParams` would ease migration for users of react + react-router.

developit:
@smeijer the trick with `useLocation().params` is that the location object returned from that hook is global to `<LocationProvider>`, so it's possible to have multiple routes (and their path patterns) simultaneously be "current":

```js
const App = () => (
  <LocationProvider>
    <ErrorBoundary>
      <div id="app">
        <header>
          <Router>
            <Profile path="/profiles/:profile" />
          </Router>
        </header>
        <Router>
          <Profile path="/profiles/:id" />
        </Router>
      </div>
    </ErrorBoundary>
  </LocationProvider>
);
```

Another option that was floated was to allow passing a path pattern to `useLocation()` itself:

```js
const { id } = useLocation('/profiles/:id').params;
```

It's more direct, but I wonder if it'd be workable?

packages/preact-iso/router.js
```diff
@@ -1,6 +1,8 @@
 import { h, createContext, cloneElement } from 'preact';
 import { useContext, useMemo, useReducer, useEffect, useRef } from 'preact/hooks';
 
+const EMPTY = {};
+
```
developit, line 5:
```suggestion
```

smeijer:
@developit, I understand. I've made a small [sandbox](https://codesandbox.io/s/react-router-useparams-of3n9?file=/App.js) to show that `react-router` doesn't care about that. Don't get me wrong, I'm not saying that this project must follow their example, I respect that this is a different project, and can have its own approach. Just sharing to show that my first expectation would be due to my experience with them.

Your example `const { id } = useLocation('/profiles/:id').params;` is definitely a nice start. It does have an impact on the developer experience tho, as we now need to connect components to url (patterns) ourselves.

As an alternative, I was thinking to simply merge all the paramters, and return the merge result instead. Your example from above would lead to `const { id, profile } = useLocation().params` in both `Profile` components, because one specified `id`, and the other `profile`. But that might lead to unexpected side effects.

Ideally, the params would be scoped to `Router` instead of `LocationProvider`.

developit:
Merging all parameters seems like a pretty reasonable default. I'll have to noodle on how to implement it though...

Another option I thought of would be to have `<Router>` re-define the `useLocation` context object. That would let each Router pass only its current `{ url, path, query, params }` object through the tree, which would support the react-router approach your sandbox uses. Maybe that's the easiest way to go.

smeijer:
The "other option" looks to me like the preferred way to go. (_to have `<Router>` re-define the `useLocation` context_). If that's the easiest one too, we have a win-win situation.(it's also what I meant with:

> Ideally, the params would be scoped to Router instead of LocationProvider.

Merging looked like the easy solution to me in case the above wouldn't be possible, but the side effects might not be worth it. Especially if the better option can be done.

I mean, merging creates the situation where one might depend on the absence of params, while another unrelated route is setting them. When developers work with generic names like `/:view/:tab/:page/:id` it's a matter of time before that happens.

JoviDeCroock:
Superseded by https://github.com/preactjs/wmr/pull/354

PR #306, opened 2021-01-15
Improve on head API

cristianbote:
Some thoughts I had for a while. As a draft so we can discuss about them.

### DIFF
--- a/src/lib/prerender.js
+++ b/src/lib/prerender.js
@@ -55,7 +55,7 @@ async function workerCode({ cwd, out, publicPath }) {
 					this.attributes.push({ name, value });
 					this[name] = value;
 				},
-				toString: function toString() {
+				get innerHTML() {
 					let attrs = this.attributes.map(([name, value]) => `${name}="${value}"`).join(' ');
 					let content = `<${this.type}`;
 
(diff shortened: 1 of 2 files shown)


src/lib/prerender.js
```diff
@@ -47,13 +47,38 @@ async function workerCode({ cwd, out, publicPath }) {
...
-			return { type };
+			const element = {
+				type,
+				attributes: [],
+				innerHTML: '',
+				setAttribute: function setAttribute(name, value) {
+					this.attributes.push([name, value]);
+					this[name] = value;
+				},
+				toString: function toString() {
```
developit, line 58:
```suggestion
				get innerHTML() {
```

src/lib/prerender.js
```diff
@@ -47,13 +47,38 @@ async function workerCode({ cwd, out, publicPath }) {
...
 		createElement(type) {
-			return { type };
+			const element = {
+				type,
+				attributes: [],
+				innerHTML: '',
+				setAttribute: function setAttribute(name, value) {
+					this.attributes.push([name, value]);
+					this[name] = value;
+				},
```
developit, line 57:
```suggestion
				setAttribute(name, value) {
					this.attributes.push({ name, value });
					this[name] = value;
				},
```

developit (commented):
a couple of comments, but as per Slack I agree emulating DOM here is a bit of a slippery slope. It'd be easier to justify "set `document.head.innerHTML` to a string and we will inject that into the template".

PR #307, opened 2021-01-15
feat: add `process.browser` for compatibility

ForsakenHarmony:
webpack provides this, some libraries use it to detect browser usage


developit (approved):
LGTM!

developit:
only nit was that we're now unnecessarily stringifying the keys, but it doesn't really matter since that gets inlined/fixed by Terser in prod.

ForsakenHarmony:
Does that impact anything other than potentially adding 4 bytes?

PR #309, opened 2021-01-19
chore: change cache path to not modify `node_modules`

ForsakenHarmony:


src/lib/npm-middleware-cache.js
```diff
@@ -193,5 +193,5 @@ async function fwrite(filename, data) {
  */
 function getCachePath({ module, version, path }, cwd) {
 	const tfPath = (path || '').replace(/\//g, '---');
-	return resolve(cwd || '.', `node_modules/${module}/.cache/${version}--${tfPath}.js`);
+	return resolve(cwd || '.', `.cache/node_modules/${module}/${version}--${tfPath}.js`);
```
developit, line 196: I've been thinking it might make sense to move `version` to be a directory so clearing old versions gets easier:

```suggestion
	return resolve(cwd || '.', `.cache/node_modules/${module}/${version}/${tfPath}.js`);
```

developit: ... alternatively, it might be nice to mirror the URLs - that brings us closer to being able to serve the cache directory as a full static application:

```suggestion
	return resolve(cwd || '.', `.cache/@npm/${module}@${version}/${tfPath}.js`);
```

JoviDeCroock (approved):

PR #320, opened 2021-01-31
use esbuild to minify the bundle

JoviDeCroock:
My suggestion would be:

- Enable ESBuild for dev (non-compress)
- Remove sucrase and use ESBuild for TypeScript types
- Keep using terser in production


src/lib/greenlet.js
```diff
@@ -0,0 +1,35 @@
+import { Worker } from 'worker_threads';
```
developit, line 1: heh - I had left this in the old PR because I was toying with moving Terser off the main thread. I wonder if that is still worth trying? We're already moving away from "one file" with the esbuild stuff, having a second Worker file for Terser wouldn't be the worst. Can still bundle it.

PR #321, opened 2021-01-31
Use swc in dev & prod

JoviDeCroock:
We can't merge this yet due to the dependency on `html` in `prefresh` https://github.com/preactjs/wmr/blob/main/packages/wmr/src/plugins/wmr/plugin.js#L75


packages/@wmr-plugins_nomodule/index.js
```diff
@@ -1,10 +1,17 @@
...
+import { transform } from '@swc/core';
+import Visitor from "@swc/core/Visitor";
 
 /** @param {import('../../types').Options} options */
 export default function (options) {
 	if (options.mode && options.mode !== 'build') return;
 	options.plugins.push(nomodulePlugin({}));
 }
+class LegacyRewriter extends Visitor {
+  visitImportDeclaration(e: CallExpression): Expression {
```
JoviDeCroock, line 10: Not fully compatible yet with the previous version of this afaik, is it the intention that we rewrite every import to `.legacy.js`?

developit: yep.

src/plugins/swc-plugin.js
```diff
@@ -0,0 +1,141 @@
...
+		console.log('transforming', filename);
+		const result = await swc.transform(code, options);
+		return result;
+	}
+});
+
+const createSwcPlugin = type => {
+	if (type === 'typescript') {
+		return swcPlugin({
+			plugin: m => {
```
JoviDeCroock, line 99: Currently this only seems to fire for the first plugin

src/plugins/swc-plugin.js
```diff
@@ -0,0 +1,141 @@
+import Visitor from '@swc/core/Visitor.js';
+import swc from '@swc/core';
+
+class JSXImportAppender extends Visitor.default {
+	visitModule(e) {
```
JoviDeCroock, line 5: I guess a better way to tackle this would be to do the latter half of this in a `visitImport` declaration.

src/wmr-middleware.js
```diff
@@ -68,14 +66,10 @@ export default function wmrMiddleware({
 			jsonPlugin(),
 			bundlePlugin({ inline: true, cwd }),
 			aliasesPlugin({ aliases, cwd: root }),
-			sucrasePlugin({
-				typescript: true,
-				sourcemap: false,
-				production: false
-			}),
+			swcPlugin('typescript'),
+			swcPlugin('jsx'),
```
developit, line 70: Doesn't this run 100% of files through TS + JSX transforms twice?

src/plugins/swc-plugin.js
```diff
@@ -0,0 +1,157 @@
...
+ * @param {object} [options]
+ * @param {boolean} [options.jsx]
+ * @param {boolean} [options.typescript]
+ * @returns {import('rollup').Plugin}
+ */
+const swcPlugin = (options = {}) => ({
+	name: 'swc',
+	async transform(code, filename) {
+		if (!options.typescript && !options.jsx) return null;
+		if (filename.includes('npm') || !/\.(jsx?|tsx?)$/.test(filename)) return null;
```
developit, line 147:
```suggestion
		if (/^[\0\b]/.test(filename) || !/\.(mjs|jsx?|tsx?)$/.test(filename)) return null;
```

src/plugins/swc-plugin.js
```diff
@@ -0,0 +1,157 @@
...
+
+		let result = { code };
+		if (options.typescript) result = await swc.transform(result.code, { ...typeScriptOptions, filename });
+		if (options.jsx) result = await swc.transform(result.code, { ...jsxOptions, filename });
```
developit, line 151:
```suggestion
		if (options.typescript) result = await swc.transform(result.code, { ...typeScriptOptions, filename });
		else if (options.jsx) result = await swc.transform(result.code, { ...jsxOptions, filename });
```

JoviDeCroock:
This PR requires a new version of `@swc/core` where we've fixed a few issues in the node version.

PR #322, opened 2021-01-31
make prefresh optional

JoviDeCroock:
This seperates wmr further away from Preact

### DIFF
--- a/src/cli.js
+++ b/src/cli.js
@@ -40,9 +40,9 @@ prog
 	.option('--sourcemap', 'Enable Source Maps')
 	.option('--profile', 'Generate build statistics')
 	.option('--reload', 'Switch off hmr and reload on file saves')
-	.option('--preact', 'Switch wmr to Preact specific mode (default: enabled)')
+	.option('--no-preact', 'Switch wmr to Preact specific mode')
 	.action(opts => {
-		opts.preact = bool(opts.preact || 'true');
+		opts.preact = !bool(opts['no-preact'] || 'false');
 		opts.optimize = !/false|0/.test(opts.compress);
 		opts.compress = bool(opts.compress);
 		if (/true/.test(process.env.PROFILE)) opts.profile = true;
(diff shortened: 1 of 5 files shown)


developit:
I wonder if it would make sense to infer this from the existence of `pkg.dependencies.preact`. Thoughts?

JoviDeCroock:
@developit we are not sure of that, they could have an empty pkg.json and stream in Preact

developit:
@JoviDeCroock right, it's not a perfect 1:1, I just meant as a less direct signal of whether to include prefresh or not. We scaffold the dependency in `create-wmr` and I would assume any template project would do so as well.

As long as the polarity is set like you have it here (default to including prefresh, `--no-preact` to disable) that seems okay. Just I was thinking it would be nice if it was a bit less explicit, since we don't have any other flags people would normally have to think about passing.

JoviDeCroock:
Yeah, what would you prefer? I think the features one is something I like the most personally. The implicit disable feels a bit hard with the npm-streaming however I would understand it

src/cli.js
```diff
@@ -40,7 +40,9 @@ prog
 	.option('--sourcemap', 'Enable Source Maps')
 	.option('--profile', 'Generate build statistics')
 	.option('--reload', 'Switch off hmr and reload on file saves')
+	.option('--no-preact', 'Switch wmr to Preact specific mode')
```
developit, line 43:
```suggestion
	.option('--no-preact', 'Disable Preact-specific features')
```

developit:
I think my preference is for the `features` object as well - it means it would be easy to make a plugin that disables prefresh and changes other options (jsx pragma, etc).

src/wmr-middleware.js
```diff
@@ -53,7 +53,8 @@ export default function wmrMiddleware({
 	aliases,
 	onError = NOOP,
 	onChange = NOOP,
-	plugins
+	plugins,
+	features = { preact: true }
```
developit, line 57: The defaults here should probably go in normalize-options:
https://github.com/preactjs/wmr/blob/a49b3b1505468db2af490b1ae2def93c474120f8/src/lib/normalize-options.js#L20

marvinhagemeister (approved):

PR #325, opened 2021-02-01
refactor: Adding 'preact' as a dependency to create-wmr template

rschristian:
This isn't necessarily needed, but seeing as how [`preact` is installed by `create-wmr`](https://github.com/preactjs/wmr/blob/247c7b291fbb4b684d6b3c566135ada5513ed9a3/packages/create-wmr/src/index.js#L41), I think it should probably be listed as a dependency? Especially if any inference like mentioned in https://github.com/preactjs/wmr/pull/322#issuecomment-770534992 is later added.

### DIFF
--- a/packages/create-wmr/tpl/package.json
+++ b/packages/create-wmr/tpl/package.json
@@ -13,7 +13,7 @@
 	},
 	"dependencies": {
 		"preact": "^10.5.12",
-		"preact-iso": "0.x"
+		"preact-iso": "^1.0.0"
 	},
 	"devDependencies": {
 		"wmr": "^1.2.0"
--- a/packages/create-wmr/tpl/package.json
+++ b/packages/create-wmr/tpl/package.json
@@ -12,10 +12,10 @@
 		"react-dom": "preact/compat"
 	},
 	"dependencies": {
-		"preact": "latest",
-		"preact-iso": "latest"
+		"preact": "^10.5.12",
+		"preact-iso": "0.x"
 	},
 	"devDependencies": {
-		"wmr": "latest"
+		"wmr": "^1.2.0"
 	}
 }
--- a/packages/create-wmr/tpl/package.json
+++ b/packages/create-wmr/tpl/package.json
@@ -12,10 +12,10 @@
 		"react-dom": "preact/compat"
 	},
 	"dependencies": {
-		"preact": "^10.5.12",
-		"preact-iso": "^0.2.0"
+		"preact": "latest",
+		"preact-iso": "latest"
 	},
 	"devDependencies": {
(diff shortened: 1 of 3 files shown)


developit:
@rschristian this makes sense, though I'm wondering: we should try to avoid having hard-coded versions in here since the template gets cached for a long time on users' machines. I was under the impression we were effectively doing `npm install --save preact preact-iso` as part of the template scaffolding, which would write these to the package.json - maybe I was wrong?

rschristian:
Certainly a fair point on the versioning.

As for the writing, no, I've certainly never seen it write anything to the package file. Docs for `pkg-install` seem to indicate it can or should though, will need to look into that then if that's the intended behavior.

developit:
FWIW I am (very) open to replacing `pkg-install` with something else. Ideally it'd be npm/yarn agnostic, but basically just anything that works haha.

marvinhagemeister (approved):

packages/create-wmr/tpl/package.json
```diff
@@ -12,10 +12,10 @@
 		"react-dom": "preact/compat"
 	},
 	"dependencies": {
-		"preact": "^10.5.12",
-		"preact-iso": "^0.2.0"
+		"preact": "latest",
```
JoviDeCroock, line 15: imo specifying `latest` as a default for users isn't the best thing, certainly not for newcomers, my suggestion would be to get the current latest and prefix with `^` ofcourse, that could be for another PR

rschristian:
I swapped out the hard-coded versions for `latest`. `pkg-install` is always going to install the latest, so it'll match the actual install logic at least. Not a huge fan, but that might be an improvement to the current mismatch of what is initially installed not necessarily matching the `package.json` file.

I've looked around for `pkg-install` alternatives but haven't seen anything nice or maintained unfortunately. Always possible I just haven't stumbled across something great though. Might toy around with creating something myself, but that's neither here nor there.

packages/create-wmr/tpl/package.json
rschristian: Was just writing something similar.

`create-wmr`, at least at the moment, does install `latest` on project creation. So we could instead specify versions there or `latest` here, just something to keep them matched up.

rschristian: I ended up just versioning in `pkg-install` & matched that in the template's `package.json`. 

`preact-iso` (for now) gets `0.x` as I'm guessing that's better than locking them to a minor with the 0-major semver rules.

JoviDeCroock (approved):

developit (approved):

developit (approved):

PR #328, opened 2021-02-04
Fix brittle regex in optimize-graph-plugin

### LINKED ISSUE #327: Bad replacement of asset paths in wmr build
This is the issue I was _trying_ to reproduce when I found #326.

In some instances, CSS asset paths aren't being replaced in the optimize-graph plugin, causing the output JS to reference the wrong paths.

## Test case

**index.html**
```html
<!DOCTYPE html>
<html lang="en">
<body>
  <script type="module" src="/index.js"></script>
</body>
</html>
```

**index.js**
```js
import './a.css';
import './foo.js';
import './b.css';
```

**foo.js**
```js
foo('string', notAString);
```

**a.css**
[empty]

**b.css**
```css
a{}
```
(could be empty but because of #326 it needs to be different from a.css)

## Expected output

The produced index.[hash].js file should correctly reference all emitted assets

## Actual output

The CSS is combined into a single file `assets/a.92dace5d.css` but the js file looks like this:

```js
function e(e){ ...stylesheet loading... },e("/assets/a.92dace5d.css"),foo("string",notAString),e("/assets/b.000506a4.css");
```

Note the reference to b.[hash].css hasn't been correctly rewritten.

This looks like it's down to [a brittle regex](https://github.com/preactjs/wmr/blob/71b80e0a89060c6d8d2c3b72ff9dcbbff9ef12e8/src/plugins/optimize-graph-plugin.js#L482) in the optimize plugin greedily matching all of `"string",notAString),e("/assets/b.000506a4.css"` as a single string.

I can make a PR to fix this one up.

jbt:
This fixes #327 

The previous regex was too greedy and would break if it came across a function with multiple arguments where the first was a string and the last was _not_ a string.

[Here's](https://regexr.com/5lpav) what the previous regex was doing, occasionally causing paths not to be picked up.

And [here's](https://regexr.com/5lpb8) the proposed version, which makes sure it can't accidentally go past the end of a string.

I wasn't sure how best to add tests for this though, happy to try and add something based off the repro case I found in #327

### DIFF
--- a/src/plugins/optimize-graph-plugin.js
+++ b/src/plugins/optimize-graph-plugin.js
@@ -479,7 +479,7 @@ function constructAssetToChunkMap(bundle) {
  * @param {(fn: string, url: string, quote: string) => string | null | undefined} replacer Return replacement code, or `null`/`undefined` to preserve the matched call.
  */
 function replaceSimpleFunctionCall(code, replacer) {
-	return code.replace(/([a-z$_][a-z0-9$_]*)\((['"`])(.*?)\2\)/gi, (s, fn, quote, url) => {
+	return code.replace(/([a-z$_][a-z0-9$_]*)\((['"`])((?:(?!\2)[^\\]|\\.)*?)\2\)/gi, (s, fn, quote, url) => {
 		const ret = replacer(fn, url, quote);
 		return ret == null ? s : ret;
 	});
--- a/src/plugins/optimize-graph-plugin.js
+++ b/src/plugins/optimize-graph-plugin.js
@@ -479,7 +479,7 @@ function constructAssetToChunkMap(bundle) {
  * @param {(fn: string, url: string, quote: string) => string | null | undefined} replacer Return replacement code, or `null`/`undefined` to preserve the matched call.
  */
 function replaceSimpleFunctionCall(code, replacer) {
-	return code.replace(/([a-z$_][a-z0-9$_]*)\((['"`])((?:(?!\2)[^\\]|\\.)*?)\2\)/gi, (s, fn, quote, url) => {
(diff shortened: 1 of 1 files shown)


developit:
Ah! this is an excellent find. I was pondering the issue when you opened it and had yet to stumble on the cause.

src/plugins/optimize-graph-plugin.js
```diff
@@ -479,7 +479,7 @@ function constructAssetToChunkMap(bundle) {
  * @param {(fn: string, url: string, quote: string) => string | null | undefined} replacer Return replacement code, or `null`/`undefined` to preserve the matched call.
  */
 function replaceSimpleFunctionCall(code, replacer) {
-	return code.replace(/([a-z$_][a-z0-9$_]*)\((['"`])(.*?)\2\)/gi, (s, fn, quote, url) => {
+	return code.replace(/([a-z$_][a-z0-9$_]*)\((['"`])((?:(?!\2)[^\\]|\\.)*?)\2\)/gi, (s, fn, quote, url) => {
```
developit, line 482: While we're fixing this, I'd also like to add a negative lookbehind that ensure it doesn't match dotted identifiers like `document.createElement(`:

```suggestion
	return code.replace(/(?<![.\w])([a-z$][\w$]*)\((['"`])((?:(?!\2)[^\\]|\\.)*?)\2\)/gi, (s, fn, quote, url) => {
```

jbt: Ah of course, I'll add that - I assume the `a-z$` should also include `_` since that's a valid starting char for an identifier

ForsakenHarmony (approved):

src/plugins/optimize-graph-plugin.js
developit: Hah - yes. Good catch.

developit (approved):

PR #330, opened 2021-02-05
Add prerendering of <head>, support .html URLs

### LINKED ISSUE #305: use/hook into prerender to build static html?
Is there a way to use the prerender functionality to build static html that could run without javascript?

Let's say I'm using an API in dev mode to render a list with blogs. But when I publish (build) the site, I don't want to keep this list dynamic. Instead, I just want to serve a static HTML file, that would also run without javascript.

Reason I ask, is because if this would be possible, we could also call private API's during development while serving the content statically once the site has been build & published. (not exposing api keys).

To update the site, one could then simply make a CI/CD pipeline to rebuilt de site on a regular basis or by using webhooks.

developit:
Support injecting HTML into <head> during prerendering, support prerendering HTML filenames without wrapper directory.

Fixes #305.


cristianbote (approved):
👍

reznord (approved):

cristianbote:
@developit just a heads up, as I merged in `main` to be as ready as possible 😄

JoviDeCroock (approved):

JoviDeCroock (approved):

PR #332, opened 2021-02-07
Fix HMR not working with CSS on Windows

matias4205:
I fixed a bug that only affects windows devices, where some files were triggering HMR but sending old changes to the browser.

**What caused it?**
The cause was that .css and .css.js files cache keys were not being normalized when saved, but only when being removed from **WRITE_CACHE**. Making the server always return old cached CSS files cause of the "\\" Windows path separator.

**How did I solve it?**
I created the **getCacheKeyById** function at the top of **wmr-middleware.js** file and used it on js, css and cssModule transforms functions for cacheKey normalization.

I also added **cross-env** to the dev script for Windows compatibility.

[This](https://github.com/preactjs/wmr/issues/246#issuecomment-748804874) issue helped me to find out the problem.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
 		"demo": "cd demo && node --experimental-modules ../src/cli.js 2>&1",
 		"demo:prod": "cd demo && node --experimental-modules ../src/cli.js build --prerender 2>&1",
 		"demo:serve": "cd demo && node --experimental-modules ../src/cli.js serve",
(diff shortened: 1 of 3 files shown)


developit:
This looks solid. I am wondering if it would be worth having `id` just be pre-normalized to Unix paths.

matias4205:
Yea that's definitely simpler 👍

JoviDeCroock (approved):
Awesome work, thank you!

PR #336, opened 2021-02-14
Fix assets bad url on dev mode Windows

matias4205:
I fixed a bug that only affects dev mode in Windows devices, where assets were pointing to bad urls cause of Windows backslash path separator.

**What caused it?**
The cause was that url paths to assets weren't being normalized when building inline modules. Generating modules like this:
image
Pointing to paths that don't exist:
image


developit (approved):
Nice catch!

PR #337, opened 2021-02-16
Attempt to fix the router flicker

### LINKED ISSUE #221: Old and new components in DOM on url change with preact-iso Router

cristianbote:
Fixes #221.

Looking to add a test for this.


demo/public/lib/lazy.js
```diff
@@ -3,14 +3,19 @@ import { useState, useRef } from 'preact/hooks';
...
 export default function lazy(load) {
 	let p, c;
-	return props => {
+	function inner(props) {
 		if (!p) p = load().then(m => ((c = (m && m.default) || m), 1));
 		const [, update] = useState(0);
 		const r = useRef(c);
 		if (!r.current) r.current = p.then(update);
 		if (c === undefined) throw p;
+		inner._r = true;
```
JoviDeCroock, line 12: Let's give this a full name rather than `_r` (`_resolved`) and work with mangle (in the future) to better this.

cristianbote: Oh, thank you @JoviDeCroock! Great call. Will change it.

JoviDeCroock (approved):
Nice work Cristian!

cristianbote:
Based on recent findings I've included an `update(1)` in `componentDidCatch` in order to pickup on the changes. @developit  I think now this is pretty safe for nested thrown promises, cause it does not read any children static property.

Is there any reason why this would lead to bad perf?

Let me know your thoughts.

PR #338, opened 2021-02-17
test a monorepo setup

JoviDeCroock:
We can evaluate whether we want this and move to `changesets` later on for easy publishing/versioning


packages/wmr/package.json
```diff
@@ -0,0 +1,161 @@
...
+			"eslint --fix",
+			"prettier --write"
+		],
+		"*.md": [
+			"prettier --write"
+		]
+	},
+	"husky": {
+		"hooks": {
+			"pre-commit": "lint-staged"
```
marvinhagemeister, line 89: Don't have much experience with workspaces, so I might be totally wrong: Wonder if this should be in the root package.json.

JoviDeCroock: Oh it is, I still have to do cleanup, looking into the build issues with the wmr.cjs file

marvinhagemeister (approved):
Looks good to me 👍

marvinhagemeister (approved):

PR #341, opened 2021-02-18
Update publish script for monorepo setup

marvinhagemeister:


packages/wmr/src/lib/~publish.js
```diff
@@ -18,7 +17,7 @@ const normalized = {
 	repository,
 	dependencies,
 	scripts: {
-		postpack: 'mv -f .package.json package.json || true; mv -f .package-lock.json package-lock.json || true'
+		postpack: 'mv -f .package.json package.json'
```
JoviDeCroock, line 20: Do we want to copy/move yarn.lcok as well?

JoviDeCroock (approved):

PR #348, opened 2021-02-20
ci: Revising workflow files

rschristian:
The Compressed-Size workflow had an incorrect pattern that needed to be fixed. 

This also limits the WMR build/lint/test & Compressed-Size workflows to only run when appropriate. As of now, they only are concerned about content within `packages/wmr` so the workflows paths have been updated to match. They won't run when `preact-iso` or `create-wmr` have changes.

### DIFF
--- a/.github/workflows/compressed-size.yml
+++ b/.github/workflows/compressed-size.yml
@@ -1,7 +1,10 @@
 name: compressed-size
 on:
   pull_request:
-    branches: [main]
+    branches: 
+      - main
+    paths:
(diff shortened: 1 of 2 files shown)


JoviDeCroock (approved):

marvinhagemeister (approved):

developit (approved):

developit:
Glad I checked first, was about to do this (and I wouldn't have thought of the directory filter, nice touch).

PR #354, opened 2021-02-21
support route parameters and inject them into the component

JoviDeCroock:
Valuable discussion: https://github.com/preactjs/wmr/pull/300
Resolves: #298

Ideally we'd put this onto context but given our system allowing for multiple matched routes it seems safer to inject these props into the page.

An alternative solution would be to wrap the rendered routes in a context (on the Router side) and have a new useRoute hook.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
 		"arrowParens": "avoid"
 	},
 	"lint-staged": {
-		"packages/**/{src,test}/**/*.js": [
+		"packages/{src,test}/**/*.js": [
 			"eslint --fix",
 			"prettier --write"
 		],
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
 	},
 	"eslintIgnore": [
 		"*.cjs",
+		"*.ts",
 		"packages/wmr/test/fixtures/**/*.expected.*",
 		"packages/wmr/test/fixtures/*/dist",
 		"packages/wmr/test/fixtures/*/.cache"
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
 		"arrowParens": "avoid"
 	},
 	"lint-staged": {
-		"packages/{src,test}/**/*.js": [
+		"packages/**/{src,test}/**/*.js": [
 			"eslint --fix",
 			"prettier --write"
 		],
--- a/packages/preact-iso/router.js
+++ b/packages/preact-iso/router.js
@@ -18,7 +18,7 @@ const UPDATE = (state, url, push) => {
 	return url;
 };
 
-const exec = (url, route, matches) => {
+export const exec = (url, route, matches) => {
   url = url.trim('/').split('/');
   route = (route || '').trim('/').split('/');
   for (let i=0, val; i<Math.max(url.length, route.length); i++) {
@@ -110,7 +110,8 @@ export function Router(props) {
 		if (matches) {
 			return p = cloneElement(vnode, { ...m, ...matches })
 		} else {
-			return vnode.props.default ? d = cloneElement(vnode, m) : 0, undefined
+			vnode.props.default && d = cloneElement(vnode, m);
+			return undefined;
 		}
 	});
 
--- a/packages/preact-iso/router.js
+++ b/packages/preact-iso/router.js
@@ -110,8 +110,8 @@ export function Router(props) {
 		if (matches) {
 			return p = h(
 				RouteContext.Provider,
-				{ value: {  ...matches } },
-				cloneElement(vnode, { ...m, ...matches })
+				{ value: m },
+				cloneElement(vnode, m)
 			);
 		}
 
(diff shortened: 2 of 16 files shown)


packages/wmr/demo/public/lib/loc.js
```diff
@@ -5,51 +5,89 @@ const UPDATE = (state, url, push) => {
...
 	if (push === true) history.pushState(null, '', url);
 	else if (push === false) history.replaceState(null, '', url);
 	return url;
 };
 
-export function Loc(props) {
+const exec = (url, route, matches) => {
+  url = url.trim('/').split('/');
+  route = (route || '').trim('/').split('/');
+  for (let i=0, val; i<Math.max(url.length, route.length); i++) {
```
marvinhagemeister, line 24: Would love to have a few tests for this function as routing is usual a critical feature of applications.

packages/preact-iso/router.js
```diff
@@ -18,6 +18,28 @@ const UPDATE = (state, url, push) => {
 	return url;
 };
 
+const exec = (url, route, matches) => {
+  url = url.trim('/').split('/');
```
marvinhagemeister, line 22: Is this the same function as in `/lib/loc.js`?

package.json
```diff
@@ -57,7 +57,7 @@
 		"arrowParens": "avoid"
 	},
 	"lint-staged": {
-		"packages/{src,test}/**/*.js": [
+		"packages/**/{src,test}/**/*.js": [
```
marvinhagemeister, line 60: Let's move this change to another PR as it doesn't seem to be necessary for route parameters

marvinhagemeister (commented):
Left a few comments, let me.know what you think 👍

packages/preact-iso/router.js
JoviDeCroock: It is yes, this is duplicated in demo due to testing I guess, this was the case before this PR

packages/wmr/demo/public/lib/loc.js
JoviDeCroock: @marvinhagemeister I've added a few tests for base/param/optional-param routes

packages/wmr/demo/public/lib/loc.js
```diff
@@ -5,51 +5,89 @@ const UPDATE = (state, url, push) => {
...
+    let [, m, param, flag] = (route[i] || '').match(/^(\:?)(.*?)([+*?]?)$/);
+    val = url[i];
+    // segment match:
+    if (!m && param==val) continue;
```
marvinhagemeister, line 28: Nit: The file is formatted with tabs, but this change here introduces spaces for indentation.

packages/preact-iso/router.js
```diff
@@ -82,15 +104,18 @@ export function Router(props) {
...
+	let p, d, m;
+	[].concat(props.children || []).some(vnode => {
+		const matches = exec(path, vnode.props.path, m = { path, query });
+		if (matches) {
+			return p = cloneElement(vnode, { ...m, ...matches })
+		} else {
+			if (vnode.props.default) d = cloneElement(vnode, m);
+			return undefined;
+		}
+	});
```
marvinhagemeister, line 116: A function returns `undefined` by default and is only needed if we want to bail out early. Here we don't have code after this statement, so we could rewrite this section as:

```suggestion
	[].concat(props.children || []).some(vnode => {
		const matches = exec(path, vnode.props.path, m = { path, query });
		if (matches) {
			return p = cloneElement(vnode, { ...m, ...matches })
		}
		
		if (vnode.props.default) d = cloneElement(vnode, m);
	});
```

marvinhagemeister (approved):
This is great! 🙌

packages/wmr/demo/public/lib/loc.js
```diff
@@ -58,20 +96,30 @@ export function Router(props) {
...
-	return curChildren.current.concat(prevChildren.current || []);
+
+	let p, d, m;
+	[].concat(props.children || []).some(vnode => {
+		const matches = exec(path, vnode.props.path, (m = { path, query }));
+		if (matches) {
+			return (p = cloneElement(vnode, { ...m, ...matches }));
+		} else {
+			if (vnode.props.default) d = cloneElement(vnode, m);
+			return undefined;
```
JoviDeCroock, line 113:
```suggestion
			if (vnode.props.default) d = cloneElement(vnode, m);
```

packages/preact-iso/router.js
```diff
@@ -82,20 +104,28 @@ export function Router(props) {
...
-
-	curChildren.current = a.map((p, i) => cloneElement(p, { path, query }));
+	let p, d, m;
+	[].concat(props.children || []).some(vnode => {
+		const matches = exec(path, vnode.props.path, m = { path, query });
+		if (matches) {
+			return p = h(
+				RouteContext.Provider,
+				{ value: {  ...matches } },
+				cloneElement(vnode, { ...m, ...matches })
```
developit, line 114: `m` and `matches` are the same object here. also if we can avoid using object spread, that preserves Safari 11 and Edge support. Since `cloneElement()` clones props anyway, it should be safe to skip manual object cloning here altogether:

```suggestion
				{ value: m },
				cloneElement(vnode, m)
```

PR #358, opened 2021-02-22
Route flicker 2

developit:
Some crazy stuff in here, but it does work.

h/t to @cristianbote for most of the work.

PR #364, opened 2021-02-23
Bugfix:router hydration duplication

developit:
The router flicker fix from #358 and #337 didn't account for hydration, and was resulting in `lazy()` components being deopted out of hydration and their contents duplicated. This fixes that issue by only applying the VNode reorder workaround to client-side route transitions.


cristianbote (approved):
Good catch 🙂

JoviDeCroock (approved):

PR #365, opened 2021-02-23
refactor: create-wmr checks dir before installing

rschristian:
After blowing up another project with an accidental `npm init`, I remembered this bug (?). Not sure if we really want this behavior or not, so I thought I'd bring this back up. 

The implementation here is pretty much swiped from `preact-cli`. Not a huge fan of another flag, but just an idea.

### DIFF
--- a/.changeset/grumpy-pens-decide.md
+++ b/.changeset/grumpy-pens-decide.md
@@ -1,0 +1,5 @@
+---
+'create-wmr': minor
+---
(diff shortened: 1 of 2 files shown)


JoviDeCroock (approved):

developit (approved):
Yes! This is definitely what we want.

developit (approved):

PR #366, opened 2021-02-23
fix: create-wmr was missing rollup

rschristian:
I believe after aa7ffb38714687024c6d33aab9970777dd14958c Rollup needs to be listed as a devDep of `create-wmr` in order for the symlink to exist.


JoviDeCroock (approved):
Mind running yarn at the root to ensure we got this listed in our lockfile?

rschristian:
@JoviDeCroock 

I did. Same version that's used in `wmr` so it won't create another entry.

PR #368, opened 2021-02-23
preact-iso: Fix hydrate parent definition

piotr-cz:
This fixes `parent` argument definition to be compatible with [`preact.render`](https://github.com/preactjs/preact/blob/10.5.12/src/index.d.ts#L226) and [`preact.hydrate`](https://github.com/preactjs/preact/blob/10.5.12/src/index.d.ts#L231) which are called under the hood:

Use case:

```jsx
// Argument of type 'HTMLElement' is not assignable to parameter of type 'VNode<{}>'.
hydrate(<App />, document.body)
```

### DIFF
--- a/.changeset/smart-ways-rule.md
+++ b/.changeset/smart-ways-rule.md
@@ -1,0 +1,5 @@
+---
+'preact-iso': patch
+---
+
+Fix hydrate parent definition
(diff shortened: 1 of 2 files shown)


JoviDeCroock (approved):
Awesome! Thank you

Issue #369, opened 2021-02-23
HMR CSS not working on Windows 10


OngeUK:
I've set up a basic project but every time I save a CSS or Sass file, I get a console error and the CSS doesn't update:

Firefox
` Uncaught TypeError: data.changes is undefined
    handleMessage http://localhost:8080/_wmr.js:39
    connect http://localhost:8080/_wmr.js:21
    <anonymous> http://localhost:8080/_wmr.js:25
_wmr.js:39:4`

Chrome
`Uncaught TypeError: Cannot read property 'forEach' of undefined
    at WebSocket.handleMessage (_wmr.js:39)`
    
Windows 10 Home
Yarn: 1.6.0
Node: 12.16.1

PR #370, opened 2021-02-23
(wmr) - Fix non js hmr

JoviDeCroock:
We handled all files as if they were `.js/.ts` while we should handle `.css` as a different type compared and all others as a full-reload.

Fixes: #369
Fixes: #378


packages/wmr/src/wmr-middleware.js
```diff
@@ -154,9 +156,19 @@ export default function wmrMiddleware({
 		// Delete any generated CSS Modules mapping modules:
 		if (/\.module\.css$/.test(filename)) WRITE_CACHE.delete(filename + '.js');
 
-		if (!pendingChanges.size) setTimeout(flushChanges, 60);
-		const result = bubbleUpdates(filename);
-		if (!result) {
+		if (!pendingChanges.size) timeout = setTimeout(flushChanges, 60);
+
+		if (filename.endsWith('.css') || /\.s[ac]ss$/.test(filename)) {
```
developit, line 161:
```suggestion
		if (/\.(css|s[ac]ss)$/.test(filename)) {
```

packages/wmr/src/wmr-middleware.js
```diff
@@ -154,9 +156,19 @@ export default function wmrMiddleware({
...
+			WRITE_CACHE.delete(filename);
+			pendingChanges.add('/' + filename);
+		} else if (/\.[tj]sx?$/.test(filename)) {
```
developit, line 164:
```suggestion
		} else if (/\.(mjs|[tj]sx?)$/.test(filename)) {
```

developit (approved):

PR #376, opened 2021-02-26
Fix rollup config on build

matias4205:
Fixed `ENOENT: no such file or directory, scandir '../../node_modules/devcert/dist/platforms` error on build


rschristian:
Are you making sure to run `yarn` in the root of the package? This is a mono-repo, deps should (mostly) be at the top level, not within individual packages.

developit (approved):

developit:
@rschristian I don't know much about yarn - should there be symlinks set up in `packages/wmr/node_modules`?

rschristian:
@developit There's some sort of symlinks, not super well versed tbh, but it looks like whatever is there doesn't work with Node's fs per the CI. Jovi set this just the other day when setting up the monorepo in d3362a2b14174d37073a28986892e6dbccec828b .

matias4205:
> Are you making sure to run `yarn` in the root of the package? This is a mono-repo, deps should (mostly) be at the top level, not within individual packages.
Yes! You were right, I wasn´t using yarn 😅, closing this.

rschristian:
Ah, glad to hear it. 

We're updating/correcting the contributing instructions right now, hopefully should be of help.

Issue #109, opened 2020-09-30
Switch CommonJS transform to cjs-module-lexer


developit:
Now that [cjs-module-lexer](https://github.com/guybedford/cjs-module-lexer) is battle-tested (it powers CommonJS in Node as of 14.17), we should switch to it. This gives us a clean way to support CJS-->ESM transformations like we're already doing, but without the horrible regex-based implementation I cobbled together.

- [ ] inline [my forked copy](https://github.com/developit/cjs-module-lexer/blob/expose-requires-dist/lexer.js) of `cjs-module-lexer`'s JS version with `require()` tracking
- [ ] create a new `lib/transform-cjs.js` module that wraps `cjs-module-lexer` (similar to [`lib/transform-imports.js`](https://github.com/developit/wmr/blob/master/src/lib/transform-imports.js))
- [ ] replace [fast-cjs-plugin.js](https://github.com/developit/wmr/blob/master/src/plugins/fast-cjs-plugin.js) implementation with a call to the new library
- [x] add/improve CommonJS tests
    _(improved in #156)_

PR #384, opened 2021-02-28
(create-wmr) - add react template

JoviDeCroock:
This currently has a non-functioning React implementation, this because we have issues reading the exports from both `react-dom` and `react` we could potentially solve this in a few ways:

- [use es-react](https://github.com/lukejacksonn/es-react)
- [use esinstall](https://www.npmjs.com/package/esinstall)
- Commit to https://github.com/preactjs/wmr/issues/109


JoviDeCroock:
Will pick up later, we currently seem to have a bug in the `npm-plugin` that ignores most of the exports of `es-react` https://www.runpkg.com/?es-react@16.13.1/index.js

PR #385, opened 2021-02-28
(wmr) - fix duplicate css crashing

JoviDeCroock:
This prevents a crash when multiple css files are imported.

The output in this case is still off, in the css-bundle I find the following:

```css
a{text-decoration:underline;text-decoration:none}
```

Should this be considered correct?


marvinhagemeister (approved):
Good catch 👍

packages/wmr/src/plugins/optimize-graph-plugin.js
```diff
@@ -262,7 +265,7 @@ function hoistEntryCss(graph) {
 
 			for (let i = 0; i < chunk.referencedFiles.length; i++) {
 				const f = chunk.referencedFiles[i];
-				if (!isCssFilename(f)) continue;
+				if (!isCssFilename(f) || !graph.bundle[f]) continue;
```
developit, line 268: I don't think we can skip this case, because we need to replace the style() call (`graph.replaceCssImport()`). It just needs to avoid re-injecting the CSS asset.

Something like:

```diff
+		const deleted = new Set();
		const visited = new Set();
		for (let i = 1; i < toMerge.length; i++) {
			const f = toMerge[i];
			if (visited.has(f)) continue;
			visited.add(f);
			// ...
			for (let i = 0; i < chunk.referencedFiles.length; i++) {
				const f = chunk.referencedFiles[i];
-				if (!isCssFilename(f) || !graph.bundle[f]) continue;
+				if (!isCssFilename(f)) continue;
				if (cssAsset) {
-					delete graph.bundle[f];
+					deleted.add(f);
			}
			// ...
		}
+		deleted.forEach(f => {
+			delete graph.bundle[f];
+		});
```

Issue #386, opened 2021-02-28
wmr dev server crash when editing a file


BenoitZugmeyer:
When running the dev server with the `wmr` command and editing a JS file, `wmr` crashes with the following error:

```
TypeError: n.dependents.every is not a function
    at m ([...]/node_modules/wmr/wmr.cjs:2:2649396)
    at R4e.<anonymous> ([...]/node_modules/wmr/wmr.cjs:2:2649692)
    at R4e.emit (node:events:376:20)
    at R4e.emitWithAll ([...]/node_modules/wmr/wmr.cjs:2:2622867)
    at R4e._emit ([...]/node_modules/wmr/wmr.cjs:2:2623984)
    at G5e.handleEvent ([...]/node_modules/wmr/wmr.cjs:2:2613282)
    at G5e.addOrChange ([...]/node_modules/wmr/wmr.cjs:2:2612592)
    at [...]/node_modules/wmr/wmr.cjs:2:2613997
    at c ([...]/node_modules/wmr/wmr.cjs:2:2611536)
    at [...]/node_modules/wmr/wmr.cjs:2:2611821
```

I believe this comes from the changes brought by #370, [where `mod.dependents.every` has been introduced](https://github.com/preactjs/wmr/pull/370/files#diff-e5209504deaf5532ecfaf38c97bc4ecb70e7eaa6f67ea97a632f62c72e7ed385R142), but `mod.dependents` is a `Set`, not an `Array`.

If this is not the root cause, I'll gladly provide a minimal project to reproduce the issue.

PR #387, opened 2021-02-28
(wmr) - convert dependents set to array

JoviDeCroock:
Fixes https://github.com/preactjs/wmr/issues/386

### DIFF
--- a/.changeset/warm-dragons-wonder.md
+++ b/.changeset/warm-dragons-wonder.md
@@ -1,0 +1,5 @@
+---
+'wmr': patch
+---
+
+Convert Set to Array before calling .every for hmr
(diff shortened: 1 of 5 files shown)


BenoitZugmeyer (approved):
I'm not sure if I'm allowed to write reviews, but:

* the code looks fine
* I tried your patch and it fixed my issue

Thank you very much for this!

JoviDeCroock:
@BenoitZugmeyer Thank  you for testing it out, gonna write a test and then we should be good to go

marvinhagemeister (approved):

marvinhagemeister:
+1 for having a test for this

Issue #392, opened 2021-03-04
Pre-render creates dist folder names that include route query params if present in a@ref links


danielweck:
To reproduce the bug, change the `a@href` attribute in:

https://github.com/preactjs/wmr/blob/42fa94027c0adc05b1e05b364010eb39203a12e8/packages/wmr/demo/public/header.tsx#L9

to: `<a href="/about?key=val&k=v&k1=a+b&k2=c%20d#hash">About</a>`

Then, `npm run build` shows:

```
Prerendered x pages:
  /
...
  /about?key=val&k=v&k1=a+b&k2=c%20d [from /]
...
```
... and the `dist` folder contains a subfolder named `about?key=val&k=v&k1=a+b&k2=c%20d` (it should be `about`).

To verify this in the browser, I added a `setTimeout()` of 2 seconds around the `hydrate()` call, and I modified some text in the pre-rendered HTML inside the `dist` folder, so that the transition from server-prerendered pages to client-side dynamic state is obvious. The `about` page with query params fails to load at first render, but displays correctly after `hydrate`.

PS: note that the demo doesn't include a `hydrate` call, I think that's a separate error:
https://github.com/preactjs/wmr/blob/42fa94027c0adc05b1e05b364010eb39203a12e8/packages/wmr/demo/public/index.tsx#L46-L48

developit:
Good catch!

PR #393, opened 2021-03-04
(wmr) - remove search in prerender

JoviDeCroock:
Fixes https://github.com/preactjs/wmr/issues/392

### DIFF
--- a/packages/wmr/src/lib/prerender.js
+++ b/packages/wmr/src/lib/prerender.js
@@ -131,7 +131,7 @@ async function workerCode({ cwd, out, publicPath }) {
 		if (result.links) {
 			for (let url of result.links) {
 				const parsed = new URL(url, 'http://localhost');
-				url = parsed.pathname + parsed.search;
+				url = parsed.pathname;
 				// ignore external links and one's we've already picked up
 				if (seen.has(url) || parsed.origin !== 'http://localhost') continue;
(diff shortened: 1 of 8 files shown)


cristianbote (approved):

packages/wmr/src/lib/prerender.js
```diff
@@ -131,7 +131,7 @@ async function workerCode({ cwd, out, publicPath }) {
 		if (result.links) {
 			for (let url of result.links) {
 				const parsed = new URL(url, 'http://localhost');
-				url = parsed.pathname + parsed.search;
+				url = parsed.pathname;
```
cristianbote, line 134: Wondering why we had this in the first place? 🤔 did we had a use case?

danielweck: Perhaps the idea was that query params can by definition result in a completely different page (e.g. web API endpoint /get?id=about vs. /get?id=home), compared with the hash fragment part of the URL which has different semantics in a traditional non-SPA client/server context (i.e. not used for Single Page Application routing).

marvinhagemeister (approved):

packages/wmr/src/lib/prerender.js
developit: Correct!

Issue #395, opened 2021-03-05
Issues with generating source maps


piotr-cz:
1.  The `--sourcemap` option is available only for the `wmr start` command (defined [here](https://github.com/preactjs/wmr/blob/e8b588054e095bed7dd7bc01601d246cb6b6cd00/packages/wmr/src/cli.js#L40)) - but doesn't trigger source maps generation.

2.  When using `wmr build  --sourcemap` (undocumented) - source maps are generated, but following warning shows up in console:
    ```
    Sourcemap is likely to be incorrect: a plugin (wmr-styles) was used to transform files, but didn't generate a sourcemap for the transformation. Consult the plugin documentation for help
    ```

    This is probably caused by missing `map` property: in the `transform` method https://github.com/preactjs/wmr/blob/e8b588054e095bed7dd7bc01601d246cb6b6cd00/packages/wmr/src/plugins/wmr/styles-plugin.js#L175-L179

developit:
Hiya! Good catch. We don't actually support sourcemaps during dev because there is so little transformation happening. We could in prod though, that plugin should just be returning `{ map: null }`.

piotr-cz:
@developit Thank you for tips! 👍

PR #397, opened 2021-03-05
(wmr) Move sourcemap option from start to build command

piotr-cz:
Fixes issue 1 in #395


piotr-cz:
Ah, I created branch from #396 so this PR includes both fixes

developit (approved):
Nice work!

PR #400, opened 2021-03-07
(ssr) - create initial template for ssr plugin

JoviDeCroock:
Currently facing a few Windows issues but slowly getting there, this is a WIP

This is the code from https://github.com/preactjs/wmr/pull/169


packages/ssr-plugin/example/public/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+	);
+}
+
+if (typeof window !== 'undefined') {
+	if (document.querySelector('.app')) {
+		hydrate(<App />, document.body);
+	} else {
+		render(<App />, document.body);
+	}
+}
```
danielweck, line 49: Is this necessary, given that `hydrate` is a wrapper that calls `render` if `! document.querySelector('script[type=isodata]')`? (I suppose the answer is yes if `script[type=isodata]` isn't generated by the SSR method implemented in this PR, but in this case could this be clarified in a code comment, for the avoidance of doubt?)

Code reference:

https://github.com/preactjs/wmr/blob/fe51b7eb3a65740571863fbeebf3367b74a9ddfe/packages/preact-iso/hydrate.js#L8-L15

The current documentation states:

https://github.com/preactjs/wmr/blob/main/packages/preact-iso/README.md#hydratejs

> hydrate() is a thin wrapper around Preact's hydrate() method. It performs hydration when the HTML for the current page includes pre-rendered output from prerender(). It falls back to plain rendering in any other cases, which is useful if you're not pre-rendering during development. This method also checks to make sure its running in a browser context before attempting any rendering - if not, it does nothing.

packages/ssr-plugin/example/public/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+export async function prerender(data) {
+	const { prerender } = await import('preact-iso');
+	return await prerender(<App {...data} />);
+}
+
+// @ts-ignore
+if (module.hot) module.hot.accept(u => render(<u.module.App />, document.body));
```
danielweck, line 57: Isn't `prefresh` activated by default? (should this line be removed?)

Related issue: https://github.com/preactjs/wmr/issues/402

JoviDeCroock: Answerred

packages/ssr-plugin/example/public/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+					</Router>
+				</ErrorBoundary>
+			</div>
+		</LocationProvider>
+	);
+}
+
+if (typeof window !== 'undefined') {
+	hydrate(<App />, document.body);
+}
```
danielweck, line 45: You will probably see this in PR #403 ( https://github.com/preactjs/wmr/pull/403#pullrequestreview-605831104 ) but just in case this goes under the radar for this PR #400, here it is again:

The `typeof window !== 'undefined'` browser check can be removed.

References:

https://github.com/preactjs/wmr/blob/fe51b7eb3a65740571863fbeebf3367b74a9ddfe/packages/preact-iso/hydrate.js#L7

https://github.com/preactjs/wmr/blob/main/packages/preact-iso/README.md#hydratejs
> This method also checks to make sure its running in a browser context before attempting any rendering - if not, it does nothing.

Issue #401, opened 2021-03-07
packages/wmr/demo demonstrates static SSR prerender, but doesn't invoke hydrate()


danielweck:
Code culprit:

https://github.com/preactjs/wmr/blob/fe51b7eb3a65740571863fbeebf3367b74a9ddfe/packages/wmr/demo/public/index.tsx#L46-L53

Note that I posted a related PR comment here (`packages/ssr-plugin/example`):

https://github.com/preactjs/wmr/pull/400/files#r589004162

danielweck:
The fix consists in renaming `render` to `hydrate`, and in removing `import { render } from 'preact';`:

```
import { hydrate } from 'preact-iso';

 if (typeof window !== 'undefined') {
 	hydrate(<App />, document.body); 
}
```

PR #403, opened 2021-03-07
fix example, use preact-iso/hydrate

JoviDeCroock:
resolves https://github.com/preactjs/wmr/issues/401

### DIFF
--- a/packages/wmr/demo/public/index.tsx
+++ b/packages/wmr/demo/public/index.tsx
@@ -1,6 +1,6 @@
-import { h, render } from 'preact';
 import { LocationProvider, Router } from './lib/loc.js';
 import lazy, { ErrorBoundary } from './lib/lazy.js';
+import hydrate from './lib/hydrate';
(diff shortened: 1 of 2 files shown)


packages/wmr/demo/public/index.tsx
```diff
@@ -44,7 +44,7 @@ export function App() {
 }
 
 if (typeof window !== 'undefined') {
-	render(<App />, document.body);
+	hydrate(<App />, document.body);
```
danielweck, line 47: The `typeof window !== 'undefined'` browser check can be removed.

References:

https://github.com/preactjs/wmr/blob/fe51b7eb3a65740571863fbeebf3367b74a9ddfe/packages/preact-iso/hydrate.js#L7

https://github.com/preactjs/wmr/blob/main/packages/preact-iso/README.md#hydratejs
> This method also checks to make sure its running in a browser context before attempting any rendering - if not, it does nothing.

PR #410, opened 2021-03-09
fix: template gitignore

rschristian:
Apparently `.gitignore` files cannot be published to packages https://github.com/npm/npm/issues/7252#issuecomment-253339460. Who knew?

On project creation, currently users will get an `.npmignore` in place of their `.gitignore`. Not quite as useful. 

I opted for a post-scaffold renaming. Seemed the simplest solution, though if that's too likely to cause confusion for contributors we could just as easily swap that out with a `writeFile` and generate it entirely.


marvinhagemeister (approved):

packages/create-wmr/src/index.js
```diff
@@ -99,7 +99,9 @@ sade('create-wmr [dir]', true)
 	.parse(process.argv);
 
 async function scaffold({ cwd, fields }) {
-	await templateDir(resolve(__dirname, '../tpl'), resolve(cwd || '.'), fields);
+	const to = resolve(cwd || '.');
+	await templateDir(resolve(__dirname, '../tpl'), to, fields);
+	await fs.rename(resolve(to, 'gitignore'), resolve(to, '.gitignore'));
```
JoviDeCroock, line 104: Shouldn't this rename from .npmignore?

rschristian: That could work, but I renamed `.gitignore` to `gitignore`. Could revert that, and just rename the `.npmignore`. 

Actually that's probably a better solution. That way, in the repo, it at least looks right.

rschristian: Updated. I do like that more.

packages/create-wmr/src/index.js
```diff
@@ -99,7 +99,11 @@ sade('create-wmr [dir]', true)
 	.parse(process.argv);
 
 async function scaffold({ cwd, fields }) {
-	await templateDir(resolve(__dirname, '../tpl'), resolve(cwd || '.'), fields);
+	const to = resolve(cwd || '.');
+	await templateDir(resolve(__dirname, '../tpl'), to, fields);
+	// Publishing to npm renames the .gitignore to .npmignore
+	// https://github.com/npm/npm/issues/7252#issuecomment-253339460
+	await fs.rename(resolve(to, '.npmignore'), resolve(to, '.gitignore'));
```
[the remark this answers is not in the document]
rschristian: The only concern there would be the look from the repo. Wondering if that'll end up with a bunch of PRs/issues to "fix" the file name

Will do though

JoviDeCroock (approved):

PR #413, opened 2021-03-09
(preact-iso): Use history.replaceState for route function

piotr-cz:
Do not merge yet, work in progress

At this moment when using `useLocation().route('/foobar')` in the effect either `history.pushState` nor `history.replaceState` method is used and window location doesn't change.

This pull request changes it's behavior in that window location will be replaced.

It would be more convenient to choose desired action like in [preact/router](https://github.com/preactjs/preact-router#redirects):

```js
function route(url, replace=false)
```

and probably this what the `push` argument in https://github.com/preactjs/wmr/blob/58f1bffd108f45c1ac5759f744f484b5d6a8fcca/packages/preact-iso/router.js#L4 was meant for, however as the useReducer dispatcher accepts only two arguments, it's value cannot be set and in effect defaults to 
 `undefined` and may be set to `true` when invoked via `MouseEvent.

One way to solve it is to rewrite dispatcher is it would be possible to call `useLocation().route()` with following signatures:

```js
// Use history.pushState
route('/foobar);
route({ path: '/foobar', replace: false});

// Use history.replaceState
route({ path: '/foobar', replace: true});
```

### DIFF
--- a/packages/preact-iso/router.js
+++ b/packages/preact-iso/router.js
@@ -1,7 +1,11 @@
 import { h, createContext, cloneElement } from 'preact';
 import { useContext, useMemo, useReducer, useEffect, useLayoutEffect, useRef } from 'preact/hooks';
 
-const UPDATE = (state, url, push) => {
+const UPDATE = (state, url) => {
+	/** @type {boolean|undefined} - History state update strategy */
+	let push = undefined
+
+	// user click (Mouse event)
 	if (url && url.type === 'click') {
 		const link = url.target.closest('a[href]');
 		if (!link || link.origin != location.origin) return state;
@@ -9,8 +13,12 @@ const UPDATE = (state, url, push) => {
 		url.preventDefault();
 		push = true;
 		url = link.href.replace(location.origin, '');
+	// navigation (PopStateEvent)
 	} else if (typeof url !== 'string') {
 		url = location.pathname + location.search;
+	// manual invocation (useLocation().route)
+	} else {
+		push = false
 	}
 
 	if (push === true) history.pushState(null, '', url);


packages/preact-iso/router.d.ts
```diff
@@ -8,7 +8,7 @@ interface LocationHook {
 	url: string;
 	path: string;
 	query: Record<string, string>;
-	route: (url: string) => void;
+	route: (url: string | { url: string, replace?: boolean }) => void;
```
developit, line 11: Wondering if we can use `(url, replace)` here? That would match `preact-router`.

developit (commented):
It looks like this was already supported?

`route('/url', true)`

(admittedly, `push=true` might be a better default)

piotr-cz:
I prefer `route(url, replace)` as well, but as long as there is no better solution, let's keep this PR opened.

piotr-cz:
I've created new PR in #424 to fix the bug with location change after using `useLocation().route(path)`.

After it's resolved it will be easier to move forward with adding an option such as `route(path, replace)` either with this PR new one.

developit:
Nice - thanks for pulling that out, just merged it.

packages/preact-iso/router.js
```diff
@@ -1,16 +1,34 @@
...
+ * @param {string} state
+ * @param {MouseEvent|PopStateEvent|Object|string} url
+ * @return {string}
+ */
+const UPDATE = (state, url) => {
+	/** @type {boolean|undefined} - History state update strategy */
+	let push = undefined
+
+	// user click
+	if (url instanceof MouseEvent) {
```
developit, line 14: This is too restrictive. It will fail for cross-document events or synthetic events. I know the duck-typing that was here wasn't great from a TS perspective, but it's the best option we have for compatibility and size and we should keep it.

packages/preact-iso/router.js
```diff
@@ -1,16 +1,34 @@
...
 
 		url.preventDefault();
 		push = true;
 		url = link.href.replace(location.origin, '');
-	} else if (typeof url !== 'string') {
+	// navigation
+	} else if (url instanceof PopStateEvent) {
 		url = location.pathname + location.search;
+	// manual invocation: route({ path, replace })
+	} else if (typeof url === 'object') {
```
developit, line 26: Here's what I'd suggest, in order to preserve the duck typing (as mentioned in previous comment):

```js
const UPDATE = (state, update) => {
	/** @type {boolean|undefined} - History state update strategy */
	let push, url;

	if (!update || typeof update === 'string') {
		url = update;
		push = true;
	}
	else if (update.type === 'click') {
		// user click
		// @ts-ignore-next
		const link = update.target.closest('a[href]');
		if (!link || link.origin != location.origin) return state;

		update.preventDefault();
		url = link.pathname + link.search + link.hash;
		push = true;
	}
	else if (update.type === 'popstate') {
		// navigation
		url = location.pathname + location.search + location.hash;
	}
	else {
		// manual invocation: route(url) or route({ url, replace })
		url = update.url || update;
		push = !url.replace;
	}

	if (push === true) history.pushState(null, '', url);
	else if (push === false) history.replaceState(null, '', url);
	return url;
}
```

piotr-cz: Should we allow passing a value that evaluates to false (undefined/ null) with route method?
If so, then should it be added to history?
```js
	if (!update || typeof update === 'string') {
	//  ^
```

Are if the `update` is not string or event I'd say it must be an object?
```js
	else {
		// manual invocation: route(url) or route({ url, replace })
		url = update.url || update;
		//                  ^
```

packages/preact-iso/router.js
piotr-cz: okay

packages/preact-iso/router.js
piotr-cz: I've updated PR with code you suggested, adapting to eslint/ prettier rules

piotr-cz:
I've updated PR to use code suggested in https://github.com/preactjs/wmr/pull/413#discussion_r593948478

However I'm not happy with the `route({ url, replace })` signature (proposed by me) and would prefer `route(url, replace)` instead.
Let's not merge this PR yet, we can do better.

developit:
@piotr-cz I think we might be able to create a wrapper function in the provider that swaps the argument signature.

piotr-cz:
@developit 
Updated, thank you for suggestion!

packages/preact-iso/router.js
piotr-cz: @developit 
What do you think about my comments above?

piotr-cz:
Just one question:
Should I add changeset with _patch_ or _minor_ version bump?
As the `route` method is not documented, I guess patch is enough.

JoviDeCroock:
I'd go for patch yea

PR #416, opened 2021-03-09
(preact-iso) - trim leading and trailing slash

### LINKED ISSUE #414: preact-iso/router uses string.trim() incorrectly to remove leading and trailing slashes
**Describe the bug**

Code references:

https://github.com/preactjs/wmr/blob/0833c4ad25ceffae461d4d8f8643744cc0b7e080/packages/preact-iso/router.js#L22-L23

and:

https://github.com/preactjs/wmr/blob/0833c4ad25ceffae461d4d8f8643744cc0b7e080/packages/wmr/demo/public/lib/loc.js#L22-L23

The current code does not actually remove leading and trailing slashes in `route` and `url`, resulting in the subsequent `.split()` call generating empty strings in the array (just like between consecutive slashes).

**To Reproduce**

```
console.log('/segment1/segment2/'.trim('/'));
```
=>
`/segment1/segment2/`

...which is to be expected, see API doc:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim

Also see this REPL which was created for related issue #415 :

https://repl.it/@danielweck/PreactWMRRouterTrimSlashes

**Expected behavior**

Expected result:

`segment1/segment2`

Suggested code correction:

`STR.replace(/^\//, '').replace(/\/$/, '')`

 - OS: all
 - Browser: all
 - WMR Version: all

**Additional context**

### LINKED ISSUE #415: preact-iso/router does not correctly split route path into several segments (preact/router works fine)
**Describe the bug**

Code references:

https://github.com/preactjs/wmr/blob/0833c4ad25ceffae461d4d8f8643744cc0b7e080/packages/preact-iso/router.js#L21-L41

and:

https://github.com/preactjs/wmr/blob/0833c4ad25ceffae461d4d8f8643744cc0b7e080/packages/wmr/demo/public/lib/loc.js#L21-L41

**To Reproduce**

https://repl.it/@danielweck/PreactWMRRouterTrimSlashes

Note that the above REPL includes the `preact/router` implementation, to compare with `preact-iso/router`. See source code:

https://github.com/preactjs/preact-router/blob/5fb140e65241bdc42e06b7b7f0d66ff85af9a34e/src/util.js#L12-L51

**Expected behavior**

Multiple consecutive URL path segments should be identified and split correctly.

 - OS: all
 - Browser: all
 - WMR Version: all

**Additional context**

Related issue: #414

JoviDeCroock:
Fixes #415
Fixes #414

### DIFF
--- a/packages/preact-iso/router.js
+++ b/packages/preact-iso/router.js
@@ -19,21 +19,21 @@ const UPDATE = (state, url, push) => {
 };
 
 export const exec = (url, route, matches) => {
-	url = url.replace(/(^\/+|\/+$)/g, '').split('/');
-	route = (route || '').replace(/(^\/+|\/+$)/g, '').split('/');
+	url = url.split('/').filter(Boolean);
+	route = (route || '').split('/').filter(Boolean);
 	for (let i = 0, val; i < Math.max(url.length, route.length); i++) {
-		let [, m, param, flag] = (route[i] || "").match(/^(:?)(.*?)([+*?]?)$/);
+		let [, m, param, flag] = (route[i] || '').match(/^(:?)(.*?)([+*?]?)$/);
 		val = url[i];
 		// segment match:
 		if (!m && param == val) continue;
 		// segment mismatch / missing required field:
-		if (!m || (!val && flag != "?" && flag != "*")) return;
+		if (!m || (!val && flag != '?' && flag != '*')) return;
 		// field match:
 		matches[param] = val && decodeURIComponent(val);
 		// normal/optional field:
-		if (flag >= "?" || flag === '') continue;
+		if (flag >= '?' || flag === '') continue;
 		// rest (+/*) match:
-		matches[param] = url.slice(i).map(decodeURIComponent).join("/");
+		matches[param] = url.slice(i).map(decodeURIComponent).join('/');
 		break;
 	}
 
--- a/packages/preact-iso/router.js
+++ b/packages/preact-iso/router.js
@@ -20,7 +20,7 @@ const UPDATE = (state, url, push) => {
 
 export const exec = (url, route, matches) => {
 	url = url.split('/').filter(Boolean)
-	route = (route || '').split('/').filter(Boolean);
+	route = (route || '').trim('/').split('/');
 	for (let i = 0, val; i < Math.max(url.length, route.length); i++) {
 		let [, m, param, flag] = (route[i] || '').match(/^(:?)(.*?)([+*?]?)$/);
 		val = url[i];
--- a/packages/preact-iso/router.js
+++ b/packages/preact-iso/router.js
@@ -27,7 +27,7 @@ export const exec = (url, route, matches) => {
 		// segment match:
 		if (!m && param == val) continue;
 		// segment mismatch / missing required field:
-		if (!m || (!val && flag != "?" && flag != "*")) break;
+		if (!m || (!val && flag != "?" && flag != "*")) return;
 		// field match:
 		matches[param] = val && decodeURIComponent(val);
 		// normal/optional field:
--- a/packages/preact-iso/router.js
+++ b/packages/preact-iso/router.js
@@ -19,8 +19,8 @@ const UPDATE = (state, url, push) => {
 };
 
 export const exec = (url, route, matches) => {
-	url = url.replace(/(^\/|\/$)/g, '');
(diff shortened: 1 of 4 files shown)


packages/preact-iso/router.js
```diff
@@ -19,27 +19,28 @@ const UPDATE = (state, url, push) => {
 };
 
 export const exec = (url, route, matches) => {
-	url = url.trim('/').split('/');
-	route = (route || '').trim('/').split('/');
+	url = url.replace(/(^\/+|\/+$)/g, '').split('/');
+	route = (route || '').replace(/(^\/+|\/+$)/g, '').split('/');
```
developit, line 23: Not sure if this is smaller:

```suggestion
	url = url.split('/').filter(Boolean);
	route = (route || '').split('/').filter(Boolean);
```

packages/preact-iso/router.js
```diff
@@ -19,27 +19,28 @@ const UPDATE = (state, url, push) => {
...
 	for (let i = 0, val; i < Math.max(url.length, route.length); i++) {
-		let [, m, param, flag] = (route[i] || '').match(/^(:?)(.*?)([+*?]?)$/);
+		let [, m, param, flag] = (route[i] || "").match(/^(:?)(.*?)([+*?]?)$/);
```
developit, line 25: what's with the quotes? lol

packages/preact-iso/router.js
```diff
@@ -19,27 +19,28 @@ const UPDATE = (state, url, push) => {
...
 		val = url[i];
 		// segment match:
 		if (!m && param == val) continue;
 		// segment mismatch / missing required field:
-		if (!m || (!val && flag != '?' && flag != '*')) return;
+		if (!m || (!val && flag != "?" && flag != "*")) return;
```
JoviDeCroock, line 30:
```suggestion
		if (!m || (!val && flag != '?' && flag != '*')) return;
```

packages/preact-iso/router.js
```diff
@@ -19,27 +19,28 @@ const UPDATE = (state, url, push) => {
...
 		// field match:
 		matches[param] = val && decodeURIComponent(val);
 		// normal/optional field:
-		if (flag >= '?') continue;
+		if (flag >= "?" || flag === '') continue;
```
JoviDeCroock, line 34:
```suggestion
		if (flag >= '?' || flag === '') continue;
```

packages/preact-iso/router.js
```diff
@@ -19,27 +19,28 @@ const UPDATE = (state, url, push) => {
...
 		// rest (+/*) match:
-		matches[param] = url.slice(i).map(decodeURIComponent).join('/');
+		matches[param] = url.slice(i).map(decodeURIComponent).join("/");
```
JoviDeCroock, line 36:
```suggestion
		matches[param] = url.slice(i).map(decodeURIComponent).join('/');
```

packages/preact-iso/router.js
JoviDeCroock, line 25:
```suggestion
		let [, m, param, flag] = (route[i] || '').match(/^(:?)(.*?)([+*?]?)$/);
```

packages/preact-iso/router.js
```diff
@@ -19,8 +19,8 @@ const UPDATE = (state, url, push) => {
 };
 
 export const exec = (url, route, matches) => {
-	url = url.trim('/').split('/');
-	route = (route || '').trim('/').split('/');
+	url = url.split('/').filter(Boolean);
```
marvinhagemeister, line 22: Just checked because I was curious: The issue is that `trim()` has [no arguments](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/Trim) and only trims whitespace. So the original code was wrong to begin with.

marvinhagemeister (approved):

packages/preact-iso/router.js
```diff
@@ -31,7 +31,7 @@ export const exec = (url, route, matches) => {
 		// field match:
 		matches[param] = val && decodeURIComponent(val);
 		// normal/optional field:
-		if (flag >= '?') continue;
+		if (flag >= '?' || flag === '') continue;
```
developit, line 34: what causes `flag===''` here?

packages/preact-iso/router.js
developit: lol yeah that's my bad.

PR #417, opened 2021-03-09
ci: Fixing actions on PRs with better pathing tool

rschristian:
This should fix the issue of actions being required on PRs but never running due to the paths

Seems to be a decent action, does what it says on the tin: https://github.com/dorny/paths-filter

### DIFF
--- a/.github/workflows/compressed-size.yml
+++ b/.github/workflows/compressed-size.yml
@@ -1,10 +1,8 @@
 name: compressed-size
 on:
   pull_request:
-    branches: 
+    branches:
       - main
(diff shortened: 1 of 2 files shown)


JoviDeCroock (approved):

JoviDeCroock:
Oh it seems like the GitHub-UI really wants them to be there 😅

rschristian:
I believe that should resolve once `main` gets this fix? Not 100% on that though

JoviDeCroock:
Let's see and make a test-PR maybe 😅

PR #419, opened 2021-03-10
(chore) - run tests on windows

JoviDeCroock:


JoviDeCroock:
Seems like an issue with too much fs-interactions during tests, Windows didn't handle this well on Prefresh either 😅

PR #424, opened 2021-03-12
(preact-iso): Router: Fix change location after useLocation().route()

piotr-cz:
At this moment routers' reducer dispatcher behaves as follows:

- User action (MouseEvent):
  window location is updated via `history.pushState`

- User navigation via back/ forward (PopStateEvent):
  window location is left intact, as it's change is handled by browser

- Manual route change via `useLocation().route(path)`
  window location is left intact

This PR
- Fixes last case so `history.pushState` is used to add new entry to browsers history and in effect location changes.
- Moves `push`  from dispatcher argument to function code block, As reducer dispatchers consume only two arguments. it gave false indication that the argument may be chosen somehow.
- Makes it clear that in reality there is no way to set `push` to false and in effect use `history.replaceState`

### DIFF
--- a/packages/preact-iso/router.js
+++ b/packages/preact-iso/router.js
@@ -1,16 +1,17 @@
 import { h, createContext, cloneElement } from 'preact';
 import { useContext, useMemo, useReducer, useEffect, useLayoutEffect, useRef } from 'preact/hooks';
 
-const UPDATE = (state, url, push) => {
+const UPDATE = (state, url) => {
+	let push = true;
 	if (url && url.type === 'click') {
 		const link = url.target.closest('a[href]');
 		if (!link || link.origin != location.origin) return state;
 
 		url.preventDefault();
-		push = true;
 		url = link.href.replace(location.origin, '');
 	} else if (typeof url !== 'string') {
 		url = location.pathname + location.search;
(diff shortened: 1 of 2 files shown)


packages/preact-iso/router.js
```diff
@@ -1,16 +1,17 @@
 import { h, createContext, cloneElement } from 'preact';
 import { useContext, useMemo, useReducer, useEffect, useLayoutEffect, useRef } from 'preact/hooks';
 
-const UPDATE = (state, url, push) => {
+const UPDATE = (state, url) => {
+	let push = true;
```
JoviDeCroock, line 5: How can we get to the `else if (push === false) history.replaceState(null, '', url);` now?

piotr-cz: We cannot. however we never could (the `UPDATE` method is not exported, just the `route(url)` is). Now it's just obvious to notice.
Please read this PR description for explanation and try to get to evaluate `history.replaceState` on master branch if you can.

developit: Right, #413 will provide a way to use `push`.

developit (approved):

developit (approved):

PR #435, opened 2021-03-15
(wmr) - report js-build errors to the client

JoviDeCroock:
image (1)

### DIFF
--- a/packages/wmr/src/wmr-middleware.js
+++ b/packages/wmr/src/wmr-middleware.js
@@ -316,105 +316,126 @@ export const TRANSFORMS = {
 
 	// Handle individual JavaScript modules
 	async js({ id, file, prefix, res, cwd, out, NonRollup }) {
-		res.setHeader('Content-Type', 'application/javascript;charset=utf-8');
+		let code;
+		try {
+			res.setHeader('Content-Type', 'application/javascript;charset=utf-8');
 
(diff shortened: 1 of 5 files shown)


packages/wmr/src/wmr-middleware.js
```diff
@@ -316,105 +316,126 @@ export const TRANSFORMS = {
...
-				}
+			writeCacheFile(out, id, code);
 
-				return spec;
+			return code;
+		} catch (e) {
+			if (code && e.loc && e.loc.line) {
+				const splittedCode = code.split('\n');
+
+				e.codeFragment = `
```
marvinhagemeister, line 422: I wonder if we could reuse the code frame code from here:

https://github.com/preactjs/wmr/blob/57afc7d8395d1bdea37b8ad40fc365b03be96267/packages/wmr/src/lib/output-utils.js#L64

Should be as easy as calling that function and wrapping it in a `kl.stripColors()` which removes the ASCII color notation codes from the string

marvinhagemeister: We should probably have a consistent terminology throughout our code base: "code frame" vs "code fragment"

marvinhagemeister (approved):
Sweet 🙌

andyrichardson:
I like them apples 🚀

PR #438, opened 2021-03-17
Improve plugin API

marvinhagemeister:
This PR adds support for returning plugins or an array of plugins from our configuration file.

### DIFF
--- a/packages/wmr/src/lib/normalize-options.js
+++ b/packages/wmr/src/lib/normalize-options.js
@@ -120,7 +120,7 @@ export async function normalizeOptions(options, mode) {
 	debug('wmr:config')(options);
 
 	/**
-	 * @param {keyof WMRPlugin} name
+	 * @param {keyof import('wmr').Plugin} name
 	 */
 	const runConfigHook = async name => {
 		for (const plugin of options.plugins) {
--- a/packages/wmr/src/lib/normalize-options.js
+++ b/packages/wmr/src/lib/normalize-options.js
@@ -166,11 +166,14 @@ export async function normalizeOptions(options, mode) {
 
 /**
  * Deeply merge two config objects
- * @param {Partial<Options>} a
- * @param {Partial<Options>} b
- * @returns {Partial<Options>}
+ * @template {Record<string, any>} T
+ * @template {Record<string, any>} U
+ * @param {T} a
+ * @param {U} b
+ * @returns {T & U}
  */
 function mergeConfig(a, b) {
+	/** @type {any} */
 	const merged = { ...a };
 
 	for (const key in b) {
--- a/packages/wmr/src/lib/normalize-options.js
+++ b/packages/wmr/src/lib/normalize-options.js
@@ -111,13 +111,9 @@ export async function normalizeOptions(options, mode) {
 	// The execution order is: "pre" -> "normal" -> "post"
 	if (options.plugins) {
 		options.plugins = options.plugins.flat().sort((a, b) => {
-			if (a.enforce === b.enforce) return 0;
-			else if ((a.enforce === 'pre' && b.enforce !== 'pre') || b.enforce === 'post') {
-				return -1;
-			} else if (b.enforce === 'pre' || a.enforce === 'post') {
-				return 1;
-			}
-			return 0;
+			const aScore = a.enforce === 'post' ? 1 : a.enforce === 'pre' ? -1 : 0;
+			const bScore = b.enforce === 'post' ? 1 : b.enforce === 'pre' ? -1 : 0;
+			return aScore - bScore;
 		});
 	}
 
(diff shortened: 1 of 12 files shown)


JoviDeCroock (approved):
Nice, we should work on some documentation for this stuff! Great work

packages/wmr/types.d.ts
```diff
@@ -28,7 +38,7 @@ declare module 'wmr' {
 		aliases: Record<string, string>;
 		env: Record<string, string>;
 		middleware: Middleware[];
-		plugins: Plugin[];
+		plugins: WMRPlugin[];
```
developit, line 41: I have no idea how painful this would be, but I was wondering if we should have `Plugin` be the WMR Plugin interface, and use `RollupPlugin` to refer to base Rollup plugins.

packages/wmr/src/lib/normalize-options.js
```diff
@@ -97,14 +96,81 @@ export async function normalizeOptions(options, mode) {
...
 	}
 
+	/**
+	 * @param {keyof WMRPlugin} name
+	 */
+	const runConfigHook = name => {
+		options.plugins.forEach(plugin => {
+			if (!plugin[name]) return;
+
+			const res = plugin[name](options);
```
developit, line 129: Do you know if Vite's version of this allows async? I'm fine with this being synchronous, just not sure if they bothered.

packages/wmr/src/lib/normalize-options.js
```diff
@@ -97,14 +96,81 @@ export async function normalizeOptions(options, mode) {
...
-		if (custom.default) await custom.default(options);
-		if (custom[mode]) await custom[mode](options);
+		const fn = custom.default || custom[mode];
+		if (fn) {
+			const res = await fn(options);
+			if (res) {
+				options.plugins.push(res);
+			}
+		}
+	}
```
developit, line 106: We need to support running both `.default()` and `.[mode]()`:

```suggestion
		if (custom.default) {
			const res = await custom.default(options);
			if (res) options = mergeConfig(options, res);
		}
		if (custom[mode]) {
			const res = await custom[mode](options);
			if (res) options = mergeConfig(options, res);
		}
	}
```

packages/wmr/src/lib/normalize-options.js
marvinhagemeister: Good call. They allow everything to be async which makes sense. Will update.

packages/wmr/src/wmr-middleware.js
```diff
@@ -67,6 +67,7 @@ export default function wmrMiddleware({
 
 	const NonRollup = createPluginContainer(
 		[
+			...plugins.filter(p => p.enforce === 'pre'),
```
developit, line 70: Idea: I believe `plugins` is already sorted by `enforce` here, so we could do:

```js
const split = plugins.findIndex(p => p.enforce !== 'pre');

const NonRollup = createPluginContainer(
	[
		...plugins.slice(0, split), // "pre"
		// ... etc
		...plugins.slice(split)  // undefined/"normal", then "post"
	],
	// ... etc
);

```

packages/wmr/src/lib/normalize-options.js
```diff
@@ -97,14 +96,82 @@ export async function normalizeOptions(options, mode) {
...
+	// The execution order is: "pre" -> "normal" -> "post"
+	if (options.plugins) {
+		options.plugins = options.plugins.flat().sort((a, b) => {
+			if (a.enforce === b.enforce) return 0;
+			else if ((a.enforce === 'pre' && b.enforce !== 'pre') || b.enforce === 'post') {
+				return -1;
+			} else if (b.enforce === 'pre' || a.enforce === 'post') {
+				return 1;
+			}
+			return 0;
```
developit, line 119:
```suggestion
			const aScore = a.enforce === 'post' ? 1 : a.enforce === 'pre' ? -1 : 0;
			const bScore = b.enforce === 'post' ? 1 : b.enforce === 'pre' ? -1 : 0;
			return aScore - bScore;
```

packages/wmr/src/wmr-middleware.js
marvinhagemeister: That won't work without reassigning the `plugins` variable after each slice operation.

marvinhagemeister:
Marking as draft, because I'd like to play around more with a subscription style API.

marvinhagemeister:
Marking as ready for review again, because after pondering for the past days I feel that the PR even in its current form improves the current plugin API a lot. It doesn't solve all problems but it's a good step in the right direction and solves some of the anyong ones. Namely:

- Internally `wmr.config.js` is just another plugin
- Plugins are now composable, makes presets possible
- No need to manually mutate config any more or call `options.push` yourself. Instead we'll deep-merge the return value
- Plugins can intercept requests/resolution/etc before WMR via `enforce: pre`. On top of that there is an additional `post` phase on top of the standard one.

JoviDeCroock (approved):

PR #440, opened 2021-03-17
Expose normalizePath helper

marvinhagemeister:


JoviDeCroock (approved):

marvinhagemeister:
@developit @JoviDeCroock @rschristian I'm unsure about the name `normalizePath` despite that being the name of the function in both `vite` and `rollup`. To me it's not immediately obvious what `normalize` means in this context. With the word "normalization" in regards to paths, I think of dropping dots in paths like `foo/../../` or stuff. Basically what node's `path.normalize` does.

But both we and rollup don't do anything that fancy here. We just convert the path to posix style slashes. There is an argument to be made for future proofing the name with `normalizePath` because it can contain additional normalization stuff in the future, but I'm not sure.

What do you folks think?

rschristian:
I agree. I think `normalize` is entirely context-driven and it's not immediately obvious what kind of normalization the context here requires. There's a lot of ways paths can be "normalized".

I'd be in favour of a much tighter name, even if that means losing out on a bit of future proofing. Even good docs can't fix a vague function name.

PR #454, opened 2021-03-21
(docs) - add getting-started and plugings pages

JoviDeCroock:
Currently disabled:

- `directory-plugin`  due to `[ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only file and data URLs are supported by the default ESM loader. Received protocol 'dir:'` and not working in `serve` mode. --> I think our prefixes generally just don't work in production
- `--prerender` due to `Cannot read property '__k' of undefined`  the `update` gets called from `lazy.js` and crashes within Preact itself


marvinhagemeister:
Commits from this PR were included in #497

Issue #259, opened 2020-12-16
Typescript error for css module imports


ivancuric:
Using the default demo template, I get the following error in any `index.js` file when importing from css (modules):

```
Cannot find module './style.module.css' or its corresponding type declarations.
```

I recommend using a d.ts file for common non-js imports like CRA does:
https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/lib/react-app.d.ts

marvinhagemeister:
That's strange because we already do have such a type file here: https://github.com/preactjs/wmr/blob/master/types.d.ts

ivancuric:
What's interesting is that I don't seem to have _any_ typing available. Everything is either inferred default types or `any`.

Eg [ `About` ](https://github.com/preactjs/wmr/blob/d7faaad9f6a2af6a4fcce34669cb296979e8b2ea/demo/public/pages/about/index.js#L3) is inferred as:
```ts
const About: ({ query }: {
    query: any;
}) => any
```

developit:
Hiya - any chance you have a custom `tsconfig.json` instead of using `create-wmr` to generate the project? We do ship a d.ts file for this as marvin mentioned, and it includes ambient declarations for `*.module.css`.

There is an issue where the imported value is incorrectly typed as `string` that is fixed on master but hasn't been released.

What version of TypeScript are you using? (local and/or vscode versions)

developit:
One thing to check: I believe our ambient declarations require `moduleResolution: 'node'` in tsconfig's compilerOptions.

ivancuric:
There is a `tsconfig.json` but it was generated using `create-wmr`. The whole folder was.
```json
{
	"compileOnSave": false,
	"compilerOptions": {
		"jsx": "preserve",
		"allowJs": true,
		"checkJs": true,
		// "strict": true,
		"noEmit": true,
		"moduleResolution": "node",
		"target": "ESNext",
		"module": "esnext",
		"resolveJsonModule": true,
		"allowSyntheticDefaultImports": true,
		"downlevelIteration": true
	},
	"typeAcquisition": {
		"enable": true
	}
}
```

ivancuric:
Ok, I had `  "typescript.disableAutomaticTypeAcquisition": true,` enabled in VScode for some reason.

Now two projects using `create-wmr` are reporting different errors.

Marabyte:
I was going to open a PR but it seems the code on master will fix this issue.

[...]

PR #460, opened 2021-03-24
fix: Ensures WMR types are available to user

rschristian:
Fixes https://github.com/preactjs/wmr/issues/259#issuecomment-805316476

JetBrains tools won't automatically load WMR's types, while I think VSCode and some others do. This should ensure better tooling parity.

Solution proposed by @danielweck https://github.com/preactjs/wmr/issues/259#issuecomment-805371513

### DIFF
--- a/packages/create-wmr/tpl/tsconfig.json
+++ b/packages/create-wmr/tpl/tsconfig.json
@@ -15,7 +15,7 @@
 		"allowSyntheticDefaultImports": true,
 		"downlevelIteration": true
 	},
-	"files": ["./node_modules/wmr/index.d.ts"],
+	"include": ["node_modules/wmr/index.d.ts"],
 	"typeAcquisition": {
 		"enable": true
 	}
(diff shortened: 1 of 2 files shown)


packages/create-wmr/tpl/tsconfig.json
```diff
@@ -15,6 +15,7 @@
 		"allowSyntheticDefaultImports": true,
 		"downlevelIteration": true
 	},
+	"files": ["./node_modules/wmr/index.d.ts"],
```
marvinhagemeister, line 18: Let's use TypeScripts [`include`](https://www.typescriptlang.org/tsconfig#include) property instead. That we we can add usercode by default too.

rschristian: Ah, apologies, forgot that was the implication with `files`. Fixed!

marvinhagemeister (approved):

PR #461, opened 2021-03-24
Fix Eslint and Prettier suggestions

piotr-cz:
As in title


marvinhagemeister (approved):
Good catch, thanks for filing a PR 🙌

PR #463, opened 2021-03-27
Throw if the user requests a port and it's already taken

marvinhagemeister:
Before this PR we'd silently choose a different port, even if the user explicitely requested one via `--port` or `process.env.PORT`


JoviDeCroock (approved):

marvinhagemeister:
Not sure what lighthouse is stumbling upon. Seems unrelated to this PR.

PR #465, opened 2021-03-27
Use consistent user plugin order

marvinhagemeister:


JoviDeCroock (approved):

packages/wmr/src/bundler.js
```diff
@@ -87,14 +87,16 @@ export async function bundleProd({
 		input.push('./' + pathToPosix(relative(root, abs)));
 	});
 
+	// Plugins are pre-sorted
+	const split = plugins.findIndex(p => p.enforce !== 'post');
```
marvinhagemeister, line 91: Wait this doesn't make sense. The condition should be inverted

developit: It's supposed to be `!== 'pre'` - the index that marks the end of the "pre" items.

developit: Basically the first index of null/'normal'/'post'.

marvinhagemeister: Imo `normal` plugins should be moved to the start too. You'll usually want to intercept stuff before we'd internal plugins. It was one of the key learnings in vite

PR #469, opened 2021-03-27
Add support for aliasing directories

### LINKED ISSUE #110: Add support for multiple cwd/asset paths
Most client projects I've worked on in the past year had the following directory structure:

```txt
/my-project
  /public <-- cwd
    favicon.ico
    index.html <-- should reference /src/index.js
  /src
    /components
      foo.js
    index.js <-- main entry file
  package.json
```

_Afaik this is the default of CRA and other cli's too. Haven't checked next.js._

This structure breaks `wmr`'s current assumption that everything is inside the specified `cwd`, as the files are "scattered" across `/public` and `/src`.

Over at in the our devtools code-base I'm facing a similar scenario:

```txt
/preact-devtools
  /src
    index.ts <-- main devtools entry
  /tests
    /fixtures <-- cwd
      page1.js
      page2.js
      index.html <-- should load /src/index.ts
```

A solution to that could be to be able to specify multiple `cwd` paths that are valid to be served from. On resolution it would iterate all cwd's and check if the requested file is present.

### LINKED ISSUE #394: Pulling files from package's parent directory
First of all, thanks again for creating wmr! It's lightweight & quick, so I don't have to think twice before using it in a project :-)

## Desired outcome

I'm trying to achieve a setup where there's a small demo project inside my [library's repo](https://github.com/wokwi/fastled-monaco).

The demo project resides in `/demo`, and it needs access to files from the library. I want to use the full library name ("@wokwi/fastled-monaco") in the imports. The idea is to keep the demo code consistent with how users would eventually import the package, while using the local sources so potential contributors can make changes and see the result immediately.

## wmr build

Getting it to work with `wmr build` was fairly easy - I only had to add an alias to my `package.json` and it worked right away:

```json
  "alias": {
    "@wokwi/fastled-monaco": "./../src"
  }
```

## wmr start

The development server was a completely different story. I spent half a day stitching together pieces of examples and hacked together a custom plugin that somehow makes it works:

```js
import { promises as fs } from 'fs';
import { dirname, join, resolve } from 'path';
import { fileURLToPath } from 'url';
import typescript from 'typescript';
const { transpile } = typescript;

const __dirname = dirname(fileURLToPath(import.meta.url));
const parentSrc = resolve(join(__dirname, '..', 'src'));

function parentProjectPlugin(options) {
  options.plugins.push({
    name: 'parent-project',
      if (options.mode === 'build') {
        return;
      }
      if (id === parentSrc) {
        return '\0parent-project:/@fastled-monaco/index.ts';
      }
      if (importer && importer.startsWith('\0fastled-monaco:') && !id.endsWith('.ts')) {
        return id + '.ts';
      }
    },
    async load(id) {
      const prefix = '\0fastled-monaco:';
      if (id.startsWith(prefix)) {
        const relative = id.substr(prefix.length);
        const finalPath = resolve(parentSrc, relative);
        this.addWatchFile(finalPath);
        return transpile(await fs.readFile(finalPath, { encoding: 'utf-8' }), {
          target: 'es2020',
          module: 'es2015',
        });
      }
    },
  });
}

  parentProjectPlugin(config);
}
```

So I have a setup that works, but:

1. It's very likely to break at some point
2. I don't really understand what's going on there. For instance, why did I add the `\0` prefix? Why do I manually compile the TypeScript files instead of having them going through the sucrase plugin?
3. For some reason, the `addWatchFile()` doesn't do the trick. The page doesn't automatically refresh when I change files in the library's `src/` directory. The changes are picked up when I refresh the page, so this is still usable (but not ideal).

I don't know if this use case is common, but I'd love to have a simpler solution for this. Ideally this would just work out of the box like `wmr build` does, but some way to tell wmr to map `@wokwi/fastled-monaco` to a specific directory in disk would also be as good. 

Any pointers/hints would be much appreciated, and thanks again for making wmr!

marvinhagemeister:
This PR adds support for aliasing directories, including directories outside of the current web root. This is done via a new `fs:` prefix.

Fixes #110, fixes #394

- [ ] fix production tests
- [ ] check on Windows
- [ ] add include dir allowlist to options
- [ ] Split up PR into smaller chunks if possible


marvinhagemeister:
Replaced by #528

PR #470, opened 2021-03-27
Update Lighthouse CI start regex command

marvinhagemeister:
fingers crossed 🤞


developit (approved):
Heh. I meant to bring this up when we changed the server start output. I was a little iffy about dropping Listening for that reason, but I don't think anything outside of Lighthouse is relying on it.

marvinhagemeister:
mh, according to their docs that should be it, but it looks like there is more going on...

PR #475, opened 2021-03-27
Only search for free port once

marvinhagemeister:
We already do that in `normalizeOptions` now since #470


developit (approved):
Nice, this makes sense.

PR #476, opened 2021-03-27
Auto restart dev server on config changes

marvinhagemeister:
Whenever a config file, loaded `*.env` files or `package.json` changes, we'll restart wmr. This makes for a pretty good user experience when dabbling with plugins and allows us to get rid of the custom config loading mechanism in `alias-plugin`. The client will automatically reconnect once the server is back up and running.


packages/wmr/src/lib/environment.js
```diff
@@ -46,14 +46,16 @@ export function parseEnvFile(str) {
  * Load additional environment variables from .env files.
  * @param {string} cwd
  * @param {string[]} envFiles
+ * @param {string[]} [configWachFiles]
  * @returns {Promise<Record<string, string>>}
  */
-export async function readEnvFiles(cwd, envFiles) {
+export async function readEnvFiles(cwd, envFiles, configWachFiles) {
```
ForsakenHarmony, line 52:
```suggestion
export async function readEnvFiles(cwd, envFiles, configWatchFiles) {
```

marvinhagemeister: Good catch! 🙌

marvinhagemeister:
Converting back to draft because open connections prevent the server from shutting down and restarting.

marvinhagemeister:
Alright I think that's it 🎉

packages/wmr/src/lib/normalize-options.js
```diff
@@ -82,7 +94,11 @@ export async function normalizeOptions(options, mode) {
 
 			const fileUrl = url.pathToFileURL(configFile);
 			try {
-				custom = await eval('(x => import(x))')(fileUrl);
+				// Node <= 12.18.4 returns an empty module if we append search params to
+				// the URL.
+				const importSource = supportsSearchParams ? `(x => import(x + '?t=${Date.now()}'))` : `(x => import(x))`;
+				// TODO: Do we even need eval here?
```
developit, line 100: The eval is required in order to prevent Rollup from breaking when it bundles WMR. Without the eval it'll try to bundle the dynamic import.

packages/wmr/src/lib/normalize-options.js
```diff
@@ -82,7 +94,11 @@ export async function normalizeOptions(options, mode) {
 
 			const fileUrl = url.pathToFileURL(configFile);
 			try {
-				custom = await eval('(x => import(x))')(fileUrl);
+				// Node <= 12.18.4 returns an empty module if we append search params to
```
developit, line 97: Idea for a workaround: we could create a temporary copy of the config and import that, then delete it (like we do with TS)? Though it doesn't seem worthwhile since this only affects server auto-restart after config changes.

packages/wmr/src/lib/normalize-options.js
marvinhagemeister: Makes sense, thanks for clearing it up 👍

packages/wmr/src/lib/normalize-options.js
marvinhagemeister: I think it's fair to require a somewhat recent node version. Only node versions `<=12.18.4` are affected, so everything starting from `12.19.0` will work. The newest version in the 12.x release line is `12.21.0`.

marvinhagemeister:
Merging, since it was reviewed a few times and no blockers were mentioned.

Issue #449, opened 2021-03-18
Userland plugin doesn't work if it uses the the same file extension as a built-in plugin, with a different import prefix


danielweck:
UPDATE: the bugs still exist, but I now have found solutions to work around WMR's handling of core file extensions, such as SVG and JSON files. Here is a working `text:` plugin that does not require using alternative file extensions in order to avoid colliding with WMR's own handling (e.g. `*.jsonld` for JSON or `*.xml` for SVG):

```js
/** @type {import('wmr').Plugin} */
function textPlugin() {
    const IMPORT_PREFIX = 'text:';
    const INTERNAL_PREFIX = `\0${IMPORT_PREFIX}`;

    return {
        name: 'text-plugin',

        async resolveId(id, importer) {
            if (id[0] === '\0' || id[0] === '\b') return;

            if (id.startsWith(IMPORT_PREFIX)) {
                id = id.slice(IMPORT_PREFIX.length);
            } else return;

            // necessary for SVG file extension
            // (otherwise WMR's url: import prefix plugin creates an inline data: URL for the SVG asset)
            if (config.mode === 'start') {
                id = `${id}${'\0'}`;
            }

            // just in case the given importer has an import prefix
            if (importer) {
                importer = importer.replace(/^[\0\b]\w+:/g, '');
            }

            const resolved = await this.resolve(id, importer, { skipSelf: true });
            if (!resolved) {
                return;
            }

            // necessary for JSON file extension
            // (because WMR's json: import prefix plugin handles it by default)
            if (/^[\0\b]\w+:/g.test(resolved.id)) {
                resolved.id = resolved.id.replace(/^[\0\b]\w+:/g, '');
            }

            return `${INTERNAL_PREFIX}${resolved.id}`;
        },

        async load(id) {
            if (!id.startsWith(INTERNAL_PREFIX)) return;

            id = id.slice(INTERNAL_PREFIX.length);

            // necessary for SVG file extension
            // (otherwise WMR's url: import prefix plugin creates an inline data: URL for the SVG asset)
            if (config.mode === 'start') {
                id = id.replace(/\0$/, '');
            }

            id = path.resolve(config.cwd || '.', id);

            this.addWatchFile(id);

            const s = fs.readFileSync(id, { encoding: 'utf8' });

            return `export default \`${s}\``;
        },
    };
}
textPlugin.enforce = 'pre';
config.plugins.push(textPlugin());
```

**Describe the bug**

[...]

PR #479, opened 2021-03-28
Use the same plugin array in dev and prod

marvinhagemeister:
This PR unifies the plugin array in `wmrMiddleware` and `buildProd`. This ensures that we are more consistent between dev and prod, by injecting plugins at the same place.

~~Draft, because it's based on #478 and #477~~

### DIFF
--- a/packages/wmr/src/lib/plugins.js
+++ b/packages/wmr/src/lib/plugins.js
@@ -1,0 +1,79 @@
+import htmPlugin from '../plugins/htm-plugin.js';
+import sucrasePlugin from '../plugins/sucrase-plugin.js';
+import wmrPlugin from '../plugins/wmr/plugin.js';
+import wmrStylesPlugin from '../plugins/wmr/styles-plugin.js';
+import sassPlugin from '../plugins/sass-plugin.js';
+import npmPlugin from '../plugins/npm-plugin/index.js';
+import publicPathPlugin from '../plugins/public-path-plugin.js';
+import minifyCssPlugin from '../plugins/minify-css-plugin.js';
+import htmlEntriesPlugin from '../plugins/html-entries-plugin.js';
+import aliasesPlugin from '../plugins/aliases-plugin.js';
+import processGlobalPlugin from '../plugins/process-global-plugin.js';
+import urlPlugin from '../plugins/url-plugin.js';
+import resolveExtensionsPlugin from '../plugins/resolve-extensions-plugin.js';
+import fastCjsPlugin from '../plugins/fast-cjs-plugin.js';
+import bundlePlugin from '../plugins/bundle-plugin.js';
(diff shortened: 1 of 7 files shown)


danielweck:
I like the proposed centralized plugin sequence, with in-situ prod vs. dev checking (much easier to debug / troubleshoot)
👍 
https://github.com/preactjs/wmr/pull/479/files#diff-207fc032a9bea78bbb1ae5c12cd7b5a6fec529aeb4e983da849c36d00cffbff3

As you know I keep a close eye to plugin-related changes ( https://github.com/preactjs/wmr/issues/449#issuecomment-808466358 ). I don't think this PR will solve the infinite loop observed in prod builds when importing JSON (with or without the `json:` prefix), or solve the file extension handling bug, but I will give it a try asap to be sure :)

Do you think your plugin refactor offers a good opportunity to add a built-in `text:` plugin in WMR? ( @developit though there was one already https://github.com/preactjs/wmr/issues/449#issuecomment-808606458 )

I am quite happy to continue using mine ( https://github.com/preactjs/wmr/issues/449#issuecomment-808465192 ), but I wonder whether including one in core WMR would affect the file extension handling bug ... (to be verified)

marvinhagemeister:
> As you know I keep a close eye to plugin-related changes ( #449 (comment) ). I don't think this PR will solve the infinite loop observed in prod builds when importing JSON (with or without the json: prefix), or solve the file extension handling bug, but I will give it a try asap to be sure :)

Yeah, let's tackle things one at a time :) I see this PR more as a stepping stone to solve the those issues. It removes one area where potential differences in `dev` vs `prod` might occur, which makes debugging plugin issues easier like you said.

I haven't checked yet what's going on with the json plugin. It's next on my todo list after the current bunch of open PRs have landed. We should cut a new release soon, too 👍 

> Do you think your plugin refactor offers a good opportunity to add a built-in text: plugin in WMR? ( @developit thought there was one already #449 (comment) )

I think this is a great idea and it makes a lot of sense to ship that with `wmr` out of the box. My current focus is on making the plugin API itself more stable and squashing the issues you mentioned before adding new built-in plugins.

packages/wmr/src/lib/plugins.js
```diff
@@ -0,0 +1,79 @@
...
+	const production = mode === 'build';
+
+	return [
+		...plugins.slice(0, split),
+		production && htmlEntriesPlugin({ cwd, publicPath }),
+		externalUrlsPlugin(),
+		nodeBuiltinsPlugin({ production }),
+		urlPlugin({ inline: !production, cwd }),
+		jsonPlugin({ cwd }),
+		bundlePlugin({ cwd }),
```
developit, line 42:
```suggestion
		bundlePlugin({ inline: !production, cwd }),
```

JoviDeCroock (approved):

PR #493, opened 2021-03-29
[preact-iso] Router should ignore links with target=".."

### LINKED ISSUE #488: preact-iso router always intercepts same-origin link clicks, not possible to opt-out
To reproduce, add "local" (same origin) links to any WMR demo page, e.g.:

```html
<a href="/file.svg">
link to file
</a>

<a href="/file.svg" rel="noreferrer noopener" target="_BLANK">
link to file (target BLANK)
</a>
```

As you can see, I serve static resources on the same domain, which I do not want the router to intercept.

=> the router attempts to load the referenced resources, and fails.

My current workaround is to instantiate my own higher-precedence link handler, and to author the above links with an additional attribute (`data-wmr-skip-router`):

```html
<a href="/file.svg" data-wmr-skip-router>
link to file
</a>

<a href="/file.svg" rel="noreferrer noopener" target="_BLANK" data-wmr-skip-router>
link to file (target BLANK)
</a>
```

Here's the mount / unmount logic in my root Preact app component:

```js
useEffect(() => {
    const clickHandler = (ev) => {
        if (!ev.target) {
            return;
        }
        const linkEl = ev.target.closest('a[href]');
        if (!linkEl) {
            return;
        }

        if (linkEl.origin !== window.location.origin) {
            return;
        }

        if (ev.target.getAttribute('data-wmr-skip-router')) {

            ev.stopPropagation(); // prevents preact-iso router
            ev.preventDefault(); // prevents default linking / popup menu behaviour

            if (ev.target.getAttribute('target') === '_BLANK') {
                window.open(linkEl.href, '_blank');
            } else {
                window.location.href = linkEl.href;
            }

            return;
        }
    };

    document.addEventListener('click', clickHandler, {
        capture: true,
    });

    return () => {
        document.removeEventListener('click', clickHandler, {
            capture: true,
        });
    };
}, []);
```

`preact-router` implements some filtering logic:

https://github.com/preactjs/preact-router/blob/5fb140e65241bdc42e06b7b7f0d66ff85af9a34e/src/index.js#L84-L85

```js
	// ignore links with targets and non-path URLs
	if (!href || !href.match(/^\//g) || (target && !target.match(/^_?self$/i))) return;
```

https://github.com/preactjs/preact-router/blob/5fb140e65241bdc42e06b7b7f0d66ff85af9a34e/src/index.js#L116

```js
if (t.hasAttribute('native')) return;
```

`preact-iso/router` only filters by origin match:

https://github.com/preactjs/wmr/blob/34b0ddb97ae7dce0bf907b86176cb9fc6ac947cd/packages/preact-iso/router.js#L8

developit:
This prevents the Router's link handling from intercepting clicks on links with `<a target="anything">`. Only these cases will be intercepted:

```
<a />
<a target="self" />
<a target="_self" />
<a target="_SELF" />
```

Fixes #488.

### DIFF
--- a/packages/preact-iso/router.js
+++ b/packages/preact-iso/router.js
@@ -5,7 +5,7 @@ const UPDATE = (state, url) => {
 	let push = true;
 	if (url && url.type === 'click') {
 		const link = url.target.closest('a[href]');
-		if (!link || link.origin != location.origin) return state;
+		if (!link || link.origin != location.origin || !/^(_?self)?$/i.test(link.target)) return state;
 
 		url.preventDefault();
 		url = link.href.replace(location.origin, '');
--- a/.changeset/yellow-peas-decide.md
+++ b/.changeset/yellow-peas-decide.md
@@ -1,0 +1,5 @@
+---
+"preact-iso": patch
+---
+
+[preact-iso] Prevent the Router from intercepting clicks on links with an "external" target (`target="anything"`).
--- a/packages/preact-iso/test/router.test.js
+++ b/packages/preact-iso/test/router.test.js
@@ -207,13 +207,13 @@ describe('Router', () => {
 		expect(A).toHaveBeenCalledWith({ path: '/', query: {} }, expect.anything());
 	});
 
-	it('should ignore clicks on externally-targeted links', async () => {
-		const pushState = jest.spyOn(history, 'pushState');
-
+	describe('intercepted VS external links', () => {
 		const shouldIntercept = [null, '', '_self', 'self', '_SELF'];
 		const shouldNavigate = ['_top', '_parent', '_blank', 'custom', '_BLANK'];
 
-		let loc;
+		// prevent actual navigations (not implemented in JSDOM)
+		const clickHandler = jest.fn(e => e.preventDefault());
+
 		const Route = jest.fn(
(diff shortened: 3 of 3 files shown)


packages/preact-iso/router.js
```diff
@@ -5,7 +5,7 @@ const UPDATE = (state, url) => {
 	let push = true;
 	if (url && url.type === 'click') {
 		const link = url.target.closest('a[href]');
-		if (!link || link.origin != location.origin) return state;
+		if (!link || link.origin != location.origin || !/^(_?self)?$/i.test(link.target)) return state;
```
marvinhagemeister, line 8: Let's add a test case for this 👍

danielweck:
Sounds good. Although ... come to think of it, the static SSR prerender probably needs to be tweaked as well, otherwise the links will be discovered and their target href generated on-disk.

developit:
@danielweck good point!

packages/preact-iso/router.js
developit: Added tests for all the possible values.

JoviDeCroock (approved):
That test is superb

marvinhagemeister (approved):

PR #498, opened 2021-03-30
Bugfix: show stack trace for prerendering errors instead of crashing

developit:
Prerendering errors are Strings because they're sent from a Worker. This parses them back into Error objects and falsifies the `.stack` property so it gets pretty printed. It also adds a guard around the stack trace parsing, which was throwing an exception when we tried to give it an undefined `error.stack` (it expects a String).

### DIFF
--- a/.changeset/dull-bags-attack.md
+++ b/.changeset/dull-bags-attack.md
@@ -1,0 +1,5 @@
+---
+"wmr": patch
+---
+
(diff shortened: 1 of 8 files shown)


developit:
Builds are passing, Netlify is just looking for stuff from #497.

marvinhagemeister:
Moved the undefined check to `errorstacks` as that should be handled by the library.

marvinhagemeister (approved):

PR #503, opened 2021-03-30
Docs Design

marvinhagemeister:
Still lots of stuff to do, but opening this for others to look at while it progresses.

- [x] Fix sidebar active page animation not triggering (the dot should move)
- [x] Add dark mode switch
- [ ] Add search
- [ ] Add responsive layout
- [ ] Add home page
- [ ] Swap out "home" in header with a logo
- [ ] Add aria current page markers
- [ ] Update code syntax colors to match theme
- [x] Add back blockquote styles
- [ ] ...


docs/public/lib/use-localstorage.js
```diff
@@ -0,0 +1,18 @@
+import { useState } from 'preact/hooks';
+
+/**
+ * @type {<T>(name: string, value: T) => [T, (v: T) => void]}
+ */
+export function useLocalStorage(key, initial) {
+	const [v, setValue] = useState(() => {
+		const stored = localStorage.getItem(key);
```
JoviDeCroock, line 8: Hmm, does this work with pre-rendering? The error in netlify refers to this being undefined

docs/public/styles/index.css
```diff
@@ -74,6 +74,9 @@ html[theme='dark'] {
 
 	--header-accent-height: 0.1875rem;
 }
+html[theme='dark'] .icon {
+	filter: invert(1);
```
marvinhagemeister, line 78: We need to add this to the "prefers-color-scheme: dark" block too

developit: It's there:

https://github.com/preactjs/wmr/pull/503/files#diff-c55bbfb39ca79e1c551b60cf79f7dcb1c40fba2788af584db3e739913cbe9f38R47

marvinhagemeister: whoops, my bad 🙈

marvinhagemeister:
As discussed on slack we'll merge this, so that other PRs can begin to add more doc content while we iterate on the general layout and design.

PR #504, opened 2021-03-30
Bugfix: fix route flashing for routes that render fragments

developit:
This should be the last of the issues relating to #358 + #364.


JoviDeCroock (approved):

marvinhagemeister (approved):
This is great! 🙌

PR #508, opened 2021-03-31
(wmr): Adding bundle visualizer

rschristian:
This adds a bundle visualizer behind the `--visualize` flag for `wmr build`. Visualizer plugin used: https://github.com/btd/rollup-plugin-visualizer

When ran, a `stats.html` file will be created in the root of the project and the file will be opened in the users' default user agent. They'll get a tree graph that will show the break down of their app rendered as well as with brotli and gzip. 

Seems like the most popular visualizer for Rollup, and is [on Vite's roadmap of plugins to cover](https://vite-rollup-plugins.patak.dev/#rollup-plugin-visualizer). This does increase the bundle as a bit as the template JS is ~82kb (mostly) unminified

### DIFF
--- a/packages/wmr/rollup.config.js
+++ b/packages/wmr/rollup.config.js
@@ -109,6 +109,35 @@ const config = {
 				}
 			}
 		},
+		{
+			// This ensures the template files for rollup-plugin-visualizer are inlined
+			// rather than bundleds as fs.readFile()
+			name: 'fix-visualizer',
+			transform(code, id) {
+				if (/rollup-plugin-visualizer[/\\]plugin[/\\]build-stats\.js$/.test(id)) {
+					code = code.replace(
+						/fs\.readFile\(path\.join\(__dirname,\s*(.+?)\)\s*,\s*"utf8"\s*\)/g,
+						(str, stringifiedJoin) => {
+							const path = require('path');
+							const fs = require('fs');
+							const filePathParts = stringifiedJoin
+								.replace(/['"`]+/g, '')
+								.replace(/\$\{template\}/g, 'treemap')
+								.split(', ');
+							const filepath = path.resolve(path.dirname(id), ...filePathParts);
+							try {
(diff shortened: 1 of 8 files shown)


packages/wmr/rollup.config.js
```diff
@@ -109,6 +109,35 @@ const config = {
...
+					code = code.replace(
+						/fs\.readFile\(path\.join\(__dirname,\s*(.+?)\)\s*,\s*"utf8"\s*\)/g,
+						(str, stringifiedJoin) => {
+							const path = require('path');
+							const fs = require('fs');
+							const filePathParts = stringifiedJoin
+								.replace(/['"`]+/g, '')
+								.replace(/\$\{template\}/g, 'treemap')
+								.split(', ');
+							const filepath = path.resolve(path.dirname(id), ...filePathParts);
```
rschristian, line 127: For simplicity's sake when reviewing, these are the lines that need to be replaced: https://github.com/btd/rollup-plugin-visualizer/blob/799b5337db4fbbf8b83119b6251efe9e33c0ab69/plugin/build-stats.js#L58-L59

The `template` is a visualizer format, i.e., treemap, sunburst, or network. Treemap seems to be the most common so I just went with that.

danielweck:
Well, that's interesting / timely :)
The other day I integrated `rollup-plugin-size-snapshot` ( https://github.com/TrySound/rollup-plugin-size-snapshot ) and `rollup-plugin-analyzer` ( https://github.com/doesdev/rollup-plugin-analyzer ) in my build, but the generated stats were not very useful so I aborted (not the plugins' fault, more to do with the WMR's plugin sequence I think).
I look forward to trying this PR!

JoviDeCroock (approved):
This looks awesome :shipit:

danielweck:
Tested. Seems to represent my pre-rendered bundle(s) accurately. Thank you!

developit (commented):
Awesome!

PR #520, opened 2021-04-01
Unify error logging to console

### LINKED ISSUE #514: CodeFrame: wrong location with tab indentation
**Describe the bug**
Our current code frame doesn't take tabs vs spaces into account when creating the marker. This leads to the marker being off.

**To Reproduce**

1. Open the demo app
2. Create a syntax error in a jsx file

**Expected behavior**
Marker position should be correct.

- OS: [e.g. MacOS 11.2.3, Windows 10]
- Browser: [e.g. chrome, safari]
- Node Version: [e.g. 15.8]
- WMR Version: [e.g. 1.3.1]

**Additional context**

marvinhagemeister:
Before:

Screenshot 2021-04-02 at 01 27 25

After:

Screenshot 2021-04-02 at 01 26 44

---

Before:

Screenshot 2021-04-02 at 01 27 31

After:

Screenshot 2021-04-02 at 01 30 30

Fixes #514 .

### DIFF
--- a/packages/wmr/src/plugins/wmr/client.js
+++ b/packages/wmr/src/plugins/wmr/client.js
@@ -2,19 +2,6 @@ function log(...args) {
 	console.info('[wmr] ', ...args);
 }
 
-/**
- * Strip ansi colors from a string. Inlined from kolorist:
- * https://github.com/marvinhagemeister/kolorist/blob/87a6402ec3a99ee4714ffb3f9bb54fd1d16174bd/src/index.ts#L62-L66
- * @param {string} str
- * @returns {string}
- */
-function stripColors(str) {
-	// eslint-disable-next-line no-control-regex
-	return ('' + str).replace(/\x1b\[[0-9;]+m/g, '').replace(/\x1b\]8;;.*?\x07(.*?)\x1b\]8;;\x07/g, function (_, group) {
-		return group;
-	});
-}
-
 const strip = url => url.replace(/\?t=\d+/g, '');
 
 const resolve = url => new URL(url, location.origin).href;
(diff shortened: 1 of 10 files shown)


JoviDeCroock (approved):

developit:
nice! I wonder if we should hide the stack when not running with `DEBUG=true`?

packages/wmr/src/plugins/wmr/client.js
```diff
@@ -2,6 +2,19 @@ function log(...args) {
 	console.info('[wmr] ', ...args);
 }
 
+/**
+ * Strip ansi colors from a string. Inlined from kolorist:
+ * https://github.com/marvinhagemeister/kolorist/blob/87a6402ec3a99ee4714ffb3f9bb54fd1d16174bd/src/index.ts#L62-L66
+ * @param {string} str
+ * @returns {string}
+ */
+function stripColors(str) {
```
developit, line 11: Would it make sense to invoke this on the server instead? It seems like it could be done [here](https://github.com/preactjs/wmr/pull/520/files#diff-618ca7c6f916be9a66704d923d39b5b0c2ef91ae8636e84409ec3ade9fad0561R94).

marvinhagemeister: Good call, updated the PR 👍

marvinhagemeister:
@developit I like the idea of only showing the stack traces to users when it's actionable for them. Had a peek at the code and it seems a bit more involved to add this. We'd need a way to distinguish errors caused by user code vs errors caused by the server or plugins. That touches a few areas in our code and I'd prefer to do that in a separate PR.

PR #522, opened 2021-04-02
Allow plugins to overwrite built-in loaders

### LINKED ISSUE #449: Userland plugin doesn't work if it uses the the same file extension as a built-in plugin, with a different import prefix

marvinhagemeister:
With this change default loaders are only added if no other prefix is present already in the path id.

I've tried a few other alternatives to the changes here, but they all end up working around limitations in rollup's plugin system and make user plugins pretty verbose to write. The gist of the problem is that with custom plugins we'll have to deal with recursive resolution early on.

The way that this usually occurs is that a custom plugin strips their prefix and passes the stripped id on to `this.resolve`. The plugins resolved by that have no knowledge about the original path and just see `./foo.json` and add their prefix, because they assume that this is the full import specifier. This leads to errors like the one outlined in #449 where `my-json:./foo.json` will be turned into `my-json:json:./foo.json`. The changes in this PR ensure that `my-json:./foo.json` stays that way.

We essentially split up resolution and applying default loaders into two distinctive concerns. The plugins itself should only match when their prefix is present. The `defaultLoader` plugin ensures that "raw" specifiers like `./foo.json` are turned into `json:./foo.json` from where we can just reused the existing resolution semantics.

Fixes #449


JoviDeCroock (approved):

danielweck:
I am getting uncaught errors `ERR_HTTP_HEADERS_SENT` `Cannot set headers after they are sent to the client` since around the time this PR was merge into the `main` branch. I am not 100% this PR is responsible though. Investigating...

```
(node:59869) UnhandledPromiseRejectionWarning: Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client
    at ServerResponse.setHeader (_http_outgoing.js:558:11)
    at ServerResponse.c.writeHead (/PATH_TO/node_modules/wmr/wmr.cjs:2:2663686)
    at WZe (/PATH_TO/node_modules/wmr/wmr.cjs:2:2660743)
    at Array.<anonymous> (/PATH_TO/node_modules/wmr/wmr.cjs:2:2662442)
    at u (/PATH_TO/node_modules/wmr/wmr.cjs:2:2648010)
    at l (/PATH_TO/node_modules/wmr/wmr.cjs:2:2647978)
    at Array.<anonymous> (/PATH_TO/node_modules/wmr/wmr.cjs:2:2662478)
    at u (/PATH_TO/node_modules/wmr/wmr.cjs:2:2648010)
    at l (/PATH_TO/node_modules/wmr/wmr.cjs:2:2647978)
    at Array.<anonymous> (/PATH_TO/node_modules/wmr/wmr.cjs:2:2801453)
(node:59869) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 30)
```

danielweck:
Okay, the above exception is thrown whenever I hard-refresh the page with `CMD`+`R` (any route, lazy included).

danielweck:
I tested this PR. The main goal is indeed achieved, and as demonstrated by your integration tests, this closes #449
Well done 👍 :)

However, this PR breaks "nested absolute path import" as described here:
https://github.com/preactjs/wmr/issues/449#issuecomment-810026883

...I will be searching for a workaround for this, as it is an important feature in a codebase where relative paths such as `../../../stuff.js` are a pain. I will let you know ;)

danielweck:
Ah, interesting: the "" only occurs with JSON files, not SVG. I will post a new issue once I've narrowed this down.

```
  wmr:middleware  /~/pages/test.json -> json:~/pages/test.json file: json:/PATH_TO/public/~/pages/test.json
  middleware:npm  "preact-iso/hydrate"
404 /@json/~/pages/test.json - File not found
  at  (:-1:-1)
```

danielweck:
Okay, I found a workaround for SVG files, using the same trick I used before for SVG files :)

In the `text:` plugin's `resolveId(id, importer)`:
```js
if (config.mode === 'start') id = `${id}${'\0'}`;
```
..and in `load(id)`:
```js
if (config.mode === 'start') id = id.replace(/\0$/, '');
```

I am filing a new issue.

marvinhagemeister:
@danielweck thanks for giving the fix a go 👍 The path aliasing thing is next on my list and I think I have an idea of where things get weird.

danielweck:
I filed this issue: https://github.com/preactjs/wmr/issues/524

Issue #521, opened 2021-04-01
Blank page on syntax error after reload


marvinhagemeister:
**Describe the bug**
When the user has a syntax error in a script file and reloads the page, it will be blank. The console prints the usual import error in that case. Any file changes from then on won't trigger a reload of the browser tab.

This is caused by syntax errors blocking the runtime from being loaded in the first place. To work around that we should inject the runtime into the HTML.

**To Reproduce**

1. Open demo app `yarn demo start`
2. Create a syntax error in one of the `.js` files
3. Reload the browser window
4. Fix the error from 2.

**Expected behavior**
Browser should show reload and show the page again

**Desktop (please complete the following information):**

- OS: [e.g. MacOS 11.2.3, Windows 10]
- Browser: [e.g. chrome, safari]
- Node Version: [e.g. 15.8]
- WMR Version: [e.g. 1.3.1]

**Additional context**

developit:
I think fixing #294 would fix this.

Issue #294, opened 2021-01-03
Inject WMR script into HTML


Snugug:
Per conversation https://github.com/preactjs/wmr/issues/182#issuecomment-753618593, WMR's client script currently only gets injected when a JS file is imported. Instead, the WMR server should inject it into a `script type="module"` tag on the HTML being served.

PR #523, opened 2021-04-02
serve _wmr.js from index.html

JoviDeCroock:
Resolves: https://github.com/preactjs/wmr/issues/294
Fixes: https://github.com/preactjs/wmr/issues/521

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
 	},
 	"scripts": {
 		"wmr": "yarn workspace wmr run",
-		"demo": "yarn workspace @examples/demo dev",
+		"demo": "yarn workspace @examples/demo run",
 		"docs": "yarn workspace docs",
 		"ci": "yarn wmr build && yarn --check-files && yarn demo build:prod"
 	},
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
 	},
 	"scripts": {
 		"wmr": "yarn workspace wmr run",
-		"demo": "yarn workspace @examples/demo run",
+		"demo": "yarn workspace @examples/demo dev",
 		"docs": "yarn workspace docs",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -10,7 +10,7 @@
 	},
 	"scripts": {
 		"wmr": "yarn workspace wmr run",
-		"demo": "yarn workspace @examples/demo run",
+		"demo": "yarn workspace @examples/demo dev",
```
marvinhagemeister: Nit: This was probably included by accident into the changes. With the previous variant we can do stuff like `yarn demo dev` or `yarn demo build` to call the different package scripts.

marvinhagemeister (approved):
This is so good! Makes the dev experience so much better 🙌 💯

packages/wmr/src/wmr-middleware.js
```diff
@@ -101,6 +106,7 @@ export default function wmrMiddleware(options) {
 			pendingChanges.add('/' + filename);
 		} else if (/\.(mjs|[tj]sx?)$/.test(filename)) {
 			if (!moduleGraph.has(filename)) {
+				onChange({ reload: true });
```
JoviDeCroock, line 109: handles first-load when the module has not entered the graph yet and we fix the error

packages/wmr/src/wmr-middleware.js
```diff
@@ -67,6 +67,11 @@ export default function wmrMiddleware(options) {
 		const mod = moduleGraph.get(filename);
 		if (!mod) return false;
 
+		if (mod.hasErrored) {
+			mod.hasErrored = false;
```
JoviDeCroock, line 71: When the module is in the graph we use this

PR #526, opened 2021-04-02
Bind to localhost by default instead of ::0

developit:


marvinhagemeister (approved):

lukeed (approved):

reznord (approved):

packages/wmr/src/lib/net-utils.js
```diff
@@ -77,7 +77,7 @@ export function getServerAddresses(addr, { https = false } = {}) {
 	}
 
 	const protocol = https ? 'https:' : 'http:';
-	const host = addr.address.replace('::', 'localhost');
+	const host = addr.address.replace(/^(127\.0\.0\.1|::)$/, 'localhost');
```
jviide, line 80: Should the `::` part be `::1`? `::` is the same thing as `::0` (and `0::0` and so on) while `::1` is the IPv6 loopback address (https://en.wikipedia.org/wiki/Localhost#Name_resolution).

PR #527, opened 2021-04-03
(preact-iso) - handle cancelled renders

### LINKED ISSUE #425: hydration of pre-rendered lazy route / component triggers exception
**Describe the bug**

Exception `TypeError: e.__k is null` is raised when attempting to hydrate a pre-rendered lazy route / dynamically-imported component. The error doesn't occur when hydrating a non-lazy route first, then navigating to a lazy one.

**To Reproduce**

1) `npm init wmr lazy-hydrate-error`
2) `cd lazy-hydrate-error`
3) open `index.js, add `onError` prop e.g. : `<ErrorBoundary onError={(e) => { console.log('ErrorBoundary onError: ', e); }}>`
4) `npm run build && npm run serve`
5) load `http://192.168.1.127:8080/about` in web browser + open inspector
6) CTRL/CMD+R / F5 to hard-refresh the page
7) observe `TypeError: e.__k is null`

**Expected behavior**

No error should occur.

 - OS: all
 - Browser: all
 - WMR Version: all

**Additional context**

This behaviour is problematic when needing to leverage `onError` to handle error state, see for example this related issue: https://github.com/preactjs/wmr/issues/423

JoviDeCroock:
By removing the synchronous render work we remove a few race conditions that were assumed to never occur, in this case when a pre-rendered route was used and we navigated away from it, the pre-rendered route would come back later and re-insert itself as the first child.

Fixes #425

CC @danielweck in case you have a mo, could you test this 😅 been doing a lot of testing but our tests don't seem to agree with me, mainly wondering if I'm missing something important

### DIFF
--- a/examples/demo/public/lib/lazy.js
+++ b/examples/demo/public/lib/lazy.js
@@ -3,23 +3,23 @@ import { useState, useRef } from 'preact/hooks';
 
 export default function lazy(load) {
 	let p, c;
-	function inner(props) {
-		if (!p) p = load().then(m => ((c = (m && m.default) || m), 1));
+	return props => {
 		const [, update] = useState(0);
 		const r = useRef(c);
-		if (!r.current) r.current = p.then(update);
-		if (c === undefined) throw p;
-		return h(c, props);
-	}
-
-	return inner;
+		if (!p) p = load().then(m => (c = (m && m.default) || m));
+		if (c !== undefined) return h(c, props);
+		if (!r.current) r.current = p.then(() => update(1));
+		throw p;
+	};
 }
 
 export function ErrorBoundary(props) {
 	this.componentDidCatch = absorb;
 	return props.children;
 }
+
 function absorb(err) {
 	if (err && err.then) this.__d = true;
+	// @ts-ignore
 	else if (this.props.onError) this.props.onError(err);
 }
--- a/examples/demo/public/lib/loc.js
+++ b/examples/demo/public/lib/loc.js
@@ -122,6 +122,7 @@ export function Router(props) {
 				cur.current.url = url;
 				prevChildren.current = null;
 				curChildren.current = h(Committer, {}, h(RouteContext.Provider, { value: m }, p || d));
+				if (wasPush) scrollTo(0, 0);
 				update(0);
 				return;
 			} else {
--- a/examples/demo/public/lib/loc.js
+++ b/examples/demo/public/lib/loc.js
@@ -104,7 +104,7 @@ export function Router(props) {
 		});
 
 		return h(Committer, {}, h(RouteContext.Provider, { value: m }, p || d));
-	}, [url, path, query]);
+	}, [url]);
 
 	useLayoutEffect(() => {
 		let p = pending.current;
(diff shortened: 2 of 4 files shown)


examples/demo/public/lib/lazy.js
```diff
@@ -3,23 +3,23 @@ import { useState, useRef } from 'preact/hooks';
 
 export default function lazy(load) {
 	let p, c;
-	function inner(props) {
```
JoviDeCroock, line 6: Don't mind the `demo/public/lazy/*.js` they weren't updated yet

examples/demo/public/lib/loc.js
```diff
@@ -96,33 +104,54 @@ export function Router(props) {
...
-	}, [url]);
+		} else {
+			commit();
+		}
 
-	// Hi! Wondering what this horrid line is for? That's totally reasonable, it is gross.
-	// It prevents the old route from being remounted because it got shifted in the children Array.
-	if (reverse && this.__v && this.__v.__k) this.__v.__k.reverse();
+		return () => {
+			if (cur.current.url !== url || pending.current !== p) {
```
JoviDeCroock, line 143: Rather than synchronously checking whether the url has changed we need to do this within the effect, the url and mutative ref could be equal again in a scenario where we `prerender --> lazy-load --> switch` which would mean that we push the oldChildren back up

danielweck:
> CC @danielweck in case you have a mo, could you test this 😅 been doing a lot of testing but our tests don't seem to agree with me, mainly wondering if I'm missing something important

I tested this PR's branch.

I am afraid I can reproduce the bug consistently. The lazy-loaded component still appears below non-lazy routes, whenever I rapidly navigate from a lazy to a non-lazy route. Just to be clear: this is with production builds (`npm run build --prerender && npm run serve`) where a lazy prerendered route gets hydrated after its initial render.

Also note that if the hydration is allowed to complete (in other words: no clicking on one of the header links to interrupt the lazy route), the `ErrorBoundary` `onError` captures this exception: `TypeError: e.__k is null`. My observation has been (and still is) that navigating from the lazy route to a non-lazy component is safe **ONCE** this exception has been raised. Any navigation (or in my case, any `useState()` / "set state" hook interactions) before that point => bug. So my workaround consists in disabling the SPA router (i.e. `ev.preventDefault()` in my delegated listener which takes precedence over the router's own) until the user can safely click header links. This phase only takes a few milliseconds, at most a second, so the user experience does not degrade much. 

I hope you will get to the root of this problem :) (you guys rock, by the way ... keep up the good work, please ;)

JoviDeCroock:
@danielweck this approach doesn't use `__k` anymore so we should be good there. I can't actually reproduce your bug currently anymore with this branch, can reproduce it on main. Gonna try some more

developit:
@JoviDeCroock I will push my take on this to a draft PR so we can compare. Needs tests.

danielweck:
@JoviDeCroock that's an interesting / informative development, I think:
https://github.com/preactjs/wmr/pull/530#issuecomment-813882327

Issue #94, opened 2020-09-21
Invalid import path for relative alias declarations


marvinhagemeister:
With a project structure like this:

```txt
/my-project
  foo.js
  index.js
  index.html
```

and an alias declaration like this:

```json
{
  "alias": {
    "foobar": "./foo.js"
  }
}
```

and an `index.js` like this:

```js
import * as foobar from "foobar";
console.log(foobar);
```

WMR adds the `cwd` path twice. The final resolved path will be `/path/to/my-project/path/to/my-project/foo.js` instead of the expected `/path/to/my-project/foo.js`

This only happens when the alias starts with a `.`. If I declare it as an absolute posix path it is resolved correctly:

```json
{
  "alias": {
    "foobar": "/foo.js"
  }
}
```

marvinhagemeister:
Closing as the aliasing feature has been completely rewritten and supports aliasing directories now 🎉 See https://wmr.dev/docs/configuration#aliasing-and-path-mappings

PR #528, opened 2021-04-03
Add support for custom aliases via resolveId

### LINKED ISSUE #524: The improved plugin arch. is great, but JSON import breaks when used with "nested" plugin that resolves absolute resource path
The improved plugin / import prefix / file extension handling successfully fixed this issue: https://github.com/preactjs/wmr/issues/449

However, there remains an issue with JSON handling (SVG works fine), as originally described here: https://github.com/preactjs/wmr/issues/449#issuecomment-810026883

This present issue updates the above comment, based on the new plugin arch implemented in the current `main` branch.

So, the first `enforce=pre` plugin in my WMR config is one that allows me to code using "absolute" import paths (i.e. relative to the `public` folder in my case) in order to avoid using long relative paths from deep-nested TS/JS files (e.g. `import { foo } from '../../../bar.js';`).

Side note - I use the following TypeScript aliases in my `tsconfig.json` in order to support the syntax `~/PATH/TO/STUFF` in Visual Studio Code (e.g. click-to-navigate):

```json
		"baseUrl": ".",
		"paths": {
			"~/*": ["./public/*"],
		}
```

Here's the plugin:

```js
	/** @type {import('wmr').Plugin} */
	function rootPlugin() {
		return {
			name: 'root-plugin',
			async resolveId(id, _importer) {
				if (!id.startsWith('~/')) return;
				const absolute = path.join(config.cwd, id.substring(2)).replace(/\\/g, '/');
				console.log('root-plugin', absolute);
				return this.resolve(absolute);
			},
		};
	}
	rootPlugin.enforce = 'pre';
	config.plugins.push(rootPlugin());
```

This breaks with WMR's built-in `url:` plugin (works fine with `json:` plugin), when using the import prefixes **explicitly**, and **only** when referencing JSON resources (not SVG). For example, this works fine:

```js
import jsonUrl from '~/path/to/test.json';

import svgAssUrl from '~/assets/img.svg';
import svgAssUrl_ from 'url:~/assets/img.svg';

import imgUrl from 'url:~/resources/image.png';
import imgUrl_ from '~/resources/image.png';
```

...but this breaks:

```js
import jsonUrl from 'url:~/path/to/test.json';
```

The `~/` plugin works perfectly fine in combination with my `text:` plugin, both with the development server (`npm start`) and the production `dist` (`npm run build --prerender && npm run serve`), but ONLY for SVG resources, not JSON. For example:

```js
import svgCake from 'text:~/vectors/cake.svg';

// THIS CRASHES (unless I use the workaround described below)
import jsonTest from 'text:~/data/test.json';

// from here I can inline / use the textual contents of the JSON and SVG files
```

Here is the `text:` plugin with a workaround to address the JSON issue (see the `if (config.mode === 'start')` code blocks):

```js
/** @type {import('wmr').Plugin} */
function textPlugin() {
    const IMPORT_PREFIX = 'text:';
    const INTERNAL_PREFIX = `\0${IMPORT_PREFIX}`;

    return {
        name: 'text-plugin',

            if (id[0] === '\0' || id[0] === '\b') return;

            if (id.startsWith(IMPORT_PREFIX)) {
                id = id.slice(IMPORT_PREFIX.length);
            } else return;

            // necessary for JSON file extension
            // (otherwise WMR's json: import prefix plugin kicks in
            if (config.mode === 'start') {
                id = `${id}${'\0'}`;
            }

            if (!resolved) {
                return;
            }

            return `${INTERNAL_PREFIX}${resolved.id}`;
        },

        async load(id) {
            if (!id.startsWith(INTERNAL_PREFIX)) return;

            id = id.slice(INTERNAL_PREFIX.length);

            // necessary for JSON file extension
            // (otherwise WMR's json: import prefix plugin kicks in
            if (config.mode === 'start') {
                id = id.replace(/\0$/, '');
            }

            id = path.resolve(config.cwd || '.', id);

            this.addWatchFile(id);

            const s = fs.readFileSync(id, { encoding: 'utf8' });

            return `export default \`${s}\``;
        },
    };
}
textPlugin.enforce = 'pre';
config.plugins.push(textPlugin());
```

marvinhagemeister:
This PR rewrites our resolution to allow plugins to participate in id resolution if it doesn't match an actual file on disk like for `~/foo.js`. We were able to resolve `~/foo.js` before, but as soon as a prefix came into play it broke down. With this PR we are able to resolve `json:~/foo.json` too. To do that we need to go through the `resolveId()` hook first. The return value of that hook is either a file on disk or a virtual one.

Via a new `resolveFile()` function we ensure that we are allowed to load the final resolved file. Currently this is limited to the web root, but will be expanded to allow a list of directories to include from. The latter is necessary for path aliasing outside of the project root (usually done for `../src/`).

Overview of what's supported with the changes here:

- Aliasing with prefixes: `json:~/foo` -> `json:./components/foo`
- Aliasing of directories: `~` -> `./components/`
- Importing of files outside of the web root (restricted via an allowlist of directories):
  ```txt
  /project
    /public
      index.js # imports ../foo/foo.js
    /foo
      foo.js
  ```
- Same as above, but with aliases

The allowlist is stored at `options.includeDirs`. By default it includes

- Web root: `<project>/public`
- `<project>/src` if present
- Any directories resolved through aliases

Fixes #524, #110, #394, #94 .


marvinhagemeister:
Marking as draft again, because I think I found a way to improve this and make directory aliasing possible in one go.

marvinhagemeister:
Spent some more time hacking on this and I think I have a good picture now which pieces need to change. The main stuff is working and I'm able to resolve files outside of the web root and **only** from certain allowed directories. The way it works is that any import resolving outside the web root will be rewritten to a special `@path` signature that keeps the path intact.

```bash
../src/outside.js -> /@path/__/src/outside.js
```

The dots for the parent path are replaced with underscores so that the browser doesn't nuke them by resolving it to `/src/outside.js`. The same is true for the `.cache` directory.

```bash
.cache/
  __/
   src/
     outside.js
```

Once resolved, we'll serialize it back to the original notation with dots. To prevent users from loading files they shouldn't have access to the files are resolved against an allowed list of directories to import files from.

Some stuff is still missing and the PR definitely needs more tests, but I'm confident in the approach chosen here.

marvinhagemeister:
Alright, managed to get everything working, including aliasing directories. Now the PR only needs a bit of cleanup before it's ready to go. Updated the initial comment with an overview of the new user facing changes gained with this PR.

JoviDeCroock (approved):

Inviz:
Would love that includeDirs feature

marvinhagemeister:
Replaced by #552

PR #529, opened 2021-04-05
Add defineConfig helper

marvinhagemeister:
This improves intellisense in our config files. It's modeled after vite's `defineConfig`.

```js
import { defineConfig } from "wmr";

export default defineConfig({
  // your config here
});
```


developit:
I'm trying to think if there's a way to avoid publishing src/...

marvinhagemeister:
@developit Moved that file out of `src/` into the root. For now that should be all that's needed since the file is self contained 👍

packages/wmr/test/test-helpers.js
```diff
@@ -50,6 +51,18 @@ export async function teardown(env) {
 export async function loadFixture(name, env) {
 	const fixture = path.join(__dirname, 'fixtures', name);
 	await ncp(fixture, env.tmp.path);
+	// For as long as the main "wmr" entry is self contained we
```
developit, line 54: Idea: what about running `npm link` here instead?

marvinhagemeister: I'd prefer to avoid that as `npm link` relies on setting up symlinks globally. It would introduce an external point of failure in our test suite.

marvinhagemeister: This gave me an idea! We can just leverage symlinks without npm via `fs.symlink`. It seems to work locally 👍

marvinhagemeister: Tests are green on CI too 🎉

packages/wmr/index.js
```diff
@@ -0,0 +1,6 @@
+/**
+ * @type {typeof import("wmr").defineConfig}
+ */
+export function defineConfig(config) {
+	return async options => (typeof config === 'function' ? config(options) : config);
```
developit, line 5: I ran into the function-or-object thing when porting the Glitch demo from Vite to WMR, and I'm thinking maybe it makes sense for us to do this step in `normalize-options.js`, to keep `defineConfig()` as a pure pass-through for TS purposes. Thoughts?

marvinhagemeister: Good idea, that's more elegant than doing it here. `defineCinfig` should do nothing in the best case

packages/wmr/test/test-helpers.js
marvinhagemeister: It's too error prone, much easier to just copy that one file into a custom `node_modules` dir. Keeps the test fixture isolated too.

PR #533, opened 2021-04-09
fix lazy imports

codepunkt:
the `lazy` imports in `preact-iso/Readme.md` are named, which is not what is exported.


JoviDeCroock (approved):
Awesome, thank you!

PR #534, opened 2021-04-09
[preact-iso] Support manual URL via <LocationProvider url="/" />

developit:
/cc @codepunkt

### DIFF
--- a/.changeset/real-horses-promise.md
+++ b/.changeset/real-horses-promise.md
@@ -1,0 +1,5 @@
+---
+"preact-iso": patch
+---
+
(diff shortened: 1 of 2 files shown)


codepunkt:
1. Type addition for `router.d.ts`:

    ```ts
    interface LocationProviderProps {
	url?: string
    }

    export const LocationProvider: FunctionComponent<LocationProviderProps>;
    ```

2. `useMemo` and `UPDATE` also use global `location`. This is where usage without `location` available would error next.

JoviDeCroock (approved):

codepunkt:
@JoviDeCroock but it doesn't help? Confused why this was merged. 🤔

PR #535, opened 2021-04-10
(preact-iso) - attempt to allign more with the one active-child semantics of Suspense

JoviDeCroock:


developit:
Now that the docs site always uses the repo copy of preact-iso, the [deploy preview](https://deploy-preview-535--wmr-dev.netlify.app/) shows the two issues I've been trying to address via the funky workarounds and swapping logic:

1. without the `[current, previous]` children approach, navigating to parametric URLs updates the currently-mounted route in-place. Passing new data to an already-rendered route component such that it will re-suspend triggers Preact's worst-case suspense behavior: we bail out mid-update and the DOM is left in a partially mutated state. In the video below, each `/docs/:x` page is handled by the same `<DocPage>` component, and this results in the DOM for each route being appended without clearing or unmounting the previous route.
2. transitioning between routes that handled by different components (like `<Home path="/" />` to `<DocPage path="/docs/:slug?" />`) works without the `[current, previous]` swapping and funky `this.__k.reverse()` hack. However, it works by unmounting the current route and its DOM tree immediately when a link is clicked, so that the page is blank while the new route loads. We could bring in `<Suspense>` from compat to show a loading spinner at least, but it doesn't really solve the underlying issue of not being able to load a route before switching to it.


https://user-images.githubusercontent.com/105127/114286491-2a5ee500-9a2d-11eb-87e0-92cfc4c9b710.mov

I do think your approach of using the effect callback is the way to go, since we run effect callbacks even when a tree suspends. I haven't checked the timing of the cancel/inner callback but as long as it fires before the invalidating render that's ideal.

JoviDeCroock:
@developit 

I don't really see the problem in 1, the route is being loaded so that ties into point 2, showing an indicator.
I guess 2 could be solved with a  `Suspense`-like mehcanism (fallback render) and that would imo be ideal, as for pre-loading we could probably start a load as soon as links are hovered, ....

developit:
Right - I really think we should preserve the behavior of not blanking the screen (with or without a fallback / spinner). The Suspense-style mechanism is the only way I can think of to do that while accounting for the fact that loading / asynchrony can occur deep within a route. It's a more complex solution for sure, but that's where I was hoping to simplify by having it reuse the same solution (and same options hook) we have in place for <Suspense>.

danielweck:
> I really think we should preserve the behavior of not blanking the screen (with or without a fallback / spinner). Otherwise, there's no value in shipping this Router versus just using any of the other ones like Wouter - they all blank the screen and don't seamlessly swap to async routes.

+1 from me, for what it's worth.

To give you some context: my current hybrid setup is pure Preact (no React components) but with preact/compat Suspense+Lazy combined with WMR's own Lazy+Router+ErrorBoundary. I use WMR's static SSR build process to prerender all routes, but I also leverage this render pass to handle dynamic content "islands" ahead-of-time.

These lazy components are fetched and hydrated at runtime (this works in lazy or non-lazy routes), as triggered by Intersection Observer, or manually when the user expands a detail/summary element, etc. I suppose this fits the definition of partial or progressive hydration.

I must admit that at this stage, I have not tacked the "difficult" part yet: data fetching on backend vs. frontend, state hydration strategies ... but I will get to that soon :)
(I don't deploy a monolithic server + DB etc., it's all micro-services / JAMStack infra, and static SSR currently meets my needs)

Until now, I had been mostly focusing on another tricky technical aspect: precise extraction of critical CSS vs. post-hydration styles used elsewhere in the app, when the website effectively becomes an SPA (I don't mean "critical CSS" in terms of above -the-fold styling, but in relation to the first render of static dry HTML/CSS).

Although I use a CSS-in-JS library to perform just-in-time stylegen when running WMR's dev server (Twind, with Tailwind utility classes conventions), my solution erases the CSS-in-JS runtime in production builds, resulting in smaller JS bundles and lower computational overheads during hydration.

So, thank you very much for pushing the envelope with WMR. There are rough edges*, but it is super refreshing to see something built from the ground-up with DX + UX performance in mind (* recently I have been pulling my hair with an edge-case bug in the import graph module resolution thing in WMR's dev server, something to do with JS/JSX/TS/TSX file extension handling, very hard to reproduce ... I will file an issue if I can narrow it down)

The modern build tooling space is becoming a crowded "market" with Parcel2, Vite, SnowPack, Astro, etc. but I think I will stick with WMR for a while. Keep up the good work! 👍  :)

PR #537, opened 2021-04-10
Always run CSS through stylis to support nesting/etc

developit:
I've been using CSS Modules, which was effectively enabling this for me. I don't think we have any reason not to pass stuff through Stylis - it adds nesting support and whatnot.

### DIFF
--- a/packages/wmr/src/plugins/wmr/styles-plugin.js
+++ b/packages/wmr/src/plugins/wmr/styles-plugin.js
@@ -124,7 +124,6 @@ export default function wmrStylesPlugin({ cwd, hot, fullPath } = {}) {
 			if (isModular) {
 				source = await modularizeCss(source, idRelative, mappings, id);
 			} else {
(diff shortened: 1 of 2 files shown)


packages/wmr/src/plugins/wmr/styles-plugin.js
```diff
@@ -124,7 +123,8 @@ export default function wmrStylesPlugin({ cwd, hot, fullPath } = {}) {
 			const mappings = [];
 			if (isModular) {
 				source = await modularizeCss(source, idRelative, mappings, id);
-			} else if (isSass || isIcss) {
+			} else {
+				// } else if (isSass || isIcss) {
```
marvinhagemeister, line 127: Let's remove it if we don't need it 👍

JoviDeCroock (approved):

marvinhagemeister (approved):

PR #542, opened 2021-04-16
[docs] Fix ThemeSwitcher hydration mismatch

developit:
This fixes two things in the docs site:

1. The ThemeSwitcher was using the saved theme value from `localStorage` to render during initial page hydration, which causes a hydration mismatch when the theme is set to anything other than `"auto"` (the value we use during SSR).

    > To work around this, I added a third boolean argument to `useLocalStorage()` that causes it to return the provided `initial` value during the first render to ensure hydration matches SSR, then immediately self-update with the localStorage value if one was set.

2. It adds a little inline JS snippet to `<head>` that applies any saved theme from localStorage before CSS loads, which eliminates the flash-of-auto thing.

Preview of both changes:

https://user-images.githubusercontent.com/105127/115053287-f95b3600-9eac-11eb-9c33-617f91ac9000.mov

### DIFF
--- a/docs/public/components/theme-switcher.js
+++ b/docs/public/components/theme-switcher.js
@@ -1,4 +1,4 @@
-import { useCallback, useEffect } from 'preact/hooks';
+import { useCallback, useEffect, useRef } from 'preact/hooks';
 import { useLocalStorage } from '../lib/use-localstorage.js';
 
 /**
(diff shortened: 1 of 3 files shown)


marvinhagemeister (approved):
Ohh I love this! Using an inline script in head is clever 👍

PR #543, opened 2021-04-17
Clear up cached module.s[ac]ss module mappings

Inviz:
Currently `style.module.sass.js` and `style.module.scss.js` mapping get stale, so adding or changing class names in a sass-powered module file will not get acknowledged until full restart of wmr.

### DIFF
--- a/.changeset/twenty-onions-serve.md
+++ b/.changeset/twenty-onions-serve.md
@@ -1,0 +1,5 @@
+---
+'wmr': minor
+---
+
+Fixes an issue that caused scss/sass modules to not be properly updated when new classes were added to the file.
--- a/packages/wmr/src/wmr-middleware.js
+++ b/packages/wmr/src/wmr-middleware.js
@@ -99,7 +99,7 @@ export default function wmrMiddleware(options) {
 		filename = filename.split(sep).join(posix.sep);
 
 		// Delete any generated CSS Modules mapping modules:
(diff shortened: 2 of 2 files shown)


JoviDeCroock (approved):
Good spot 😅 mind adding a changeset by running `yarn changeset` and indicating your patch

Inviz:
I'll do it, but honestly it seems to solve only half of the problem. So what i'm seeing is that hot reloading won't work for new class names in `module.sass`, but refreshing page at least works now.

JoviDeCroock:
Oh interesting, problem is probably located [here](https://github.com/preactjs/wmr/blob/main/packages/wmr/src/plugins/wmr/client.js#L47) the update trigger comes through if you hit the part you've altered but we probably miss a case for sass here as well. If we can make a test I'd gladly look into it

Inviz:
I think that what is supposed to happen is that when `module.css` file is updated, all files that import it need to be re-evaluated. Because right now if I do something like:

```<div class={style.classThatDoesntExist}></div>``` first, that evalutes to `class={undefined}`. And then I go to `module.scss` and add a rule with `.classThatDoesntExist`, the original component needs to acquire the reference to that new class. Right now as i can see, the system just assumes that component doesnt need to be re-rendered, which is true when css rule is updated, but not added

marvinhagemeister (approved):
Good catch! Thank you so much for making a PR 👍

PR #547, opened 2021-04-25
fix: Error messages by removing the backspace character from npm prefix

rschristian:
Preface: I'm not sure of the reason this was used to begin with, so it's possible I've missed something or that this isn't the best fix. Test suite passes, behavior seems right, but beyond that, quite unsure. 

## Summary

Corrects WMR output when referencing imports. For example, this message was pulled from the slack: `... Error: Could not loadnpm/call-bind@1.0.2/./ (imported bynpm/call-bind@1.0.2/callBound.js): ...`. The backspace character in the NPM prefix (`\bnpm`) would cause the output to be shown incorrectly like this (`loadnpm/...`, `bynpm/...`).

This PR simply removes the backspace character from the prefixes and the places it is checked for.

As far as I know, `PREFIX_INTERNAL` wasn't an issue / ever shown, but due to the shared regex between the prefixes, I changed both.

### DIFF
--- a/.changeset/big-dolls-hammer.md
+++ b/.changeset/big-dolls-hammer.md
@@ -1,0 +1,5 @@
+---
+'wmr': patch
+---
+
+Corrects output when referrencing module with npm prefix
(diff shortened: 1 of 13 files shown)


marvinhagemeister (approved):
Sweet!! Finally it's gone 🙌 🙌

marvinhagemeister:
FYI: Took the liberty of pushing a commit on top that removes the remaining internal id checks for `\b`. It was only used with the npm plugin and can be removed all together with this PR 👍

PR #548, opened 2021-04-25
refactor: Swapping out premove with fs.rmdir

rschristian:
We originally added `premove` as the recursive removal option for `fs.rmdir()` didn't come out until Node 12.10 and the engine requirements were only set to >=12. 

But with #546 now upping that to 12.17, we can swap out `premove` with the native method.


marvinhagemeister (approved):
Good catch 👍

PR #551, opened 2021-04-26
Inline transform context

marvinhagemeister:


developit (approved):
Strange that it was hoisted in the first place... maybe it used to be used in a few places? Not sure.

PR #552, opened 2021-04-26
Add support for importing from directories outside of cwd (aliases)

marvinhagemeister:
This is an alternative implementation of #528 based on aliases instead of serializing `..` in paths.

TODO:
- [x] Add `<project>/src` as an alias by default
- [x] Add tests


developit:
Nice! This is almost exactly what I was hoping for! The only thing I'd like to check is if it's possible to simplify further by not having the source specifiers be relative paths, but just the plain prefixes / left-hand-side:

```js
import 'src/outside.js'
```

*But*, at the same time, reading through this and the test case makes me realize something you probably already had in mind: when the alias is the same as its local path, the specifier is a valid relative path and things like intellisense work (nice).

But then I also got to thinking: I think that exact relative import would work in both dev and prod without allowing relative paths as aliases? It would request `/src/outside.js`, which would try to load the module specifier `src/outside.js`, which would be resolved by the alias plugin to the aliased path. 

I'm not 100% sure what the tradeoffs would be between the two approaches. For aliases that are totally different paths from their name I think intellisense will require taconfig anyway, but I don't know for sure. 

(These are just my random thoughts lol)

marvinhagemeister:
Got it fully working! There is a bit of fine tuning I'd love to do before marking this as ready:

- Change directory alias syntax to match TypeScript's `foo/` -> `foo/*`. Makes it more explicit that we're dealing with a directory
- Waffle on whether we should add `options.includeDirs`. It makes a lot of sense to cache that as it's a frequent path

packages/wmr/src/wmr-middleware.js
```diff
@@ -348,33 +455,43 @@ export const TRANSFORMS = {
...
+
+								// Option 3: omit root package versions
+								spec = `/@npm/${meta.module}${meta.path ? '/' + meta.path : ''}`;
+							} else {
+								spec = `/@id/${spec}`;
+							}
 						}
 					}
 
-					const modSpec = spec.startsWith('../') ? spec.replace(/..\/g/, '') : spec.replace('./', '');
```
marvinhagemeister, line 377: TODO: Check if this can be reverted

.changeset/silver-ways-lie.md
```diff
@@ -0,0 +1,5 @@
+---
+'wmr': minor
+---
+
+Add support for aliasing directories outside of root
```
marvinhagemeister, line 5: TODO: Add a more detailed explanation

marvinhagemeister:
Alright, i think it's ready to be reviewed now. The diff looks a bit large due to a few new test cases and fixtures being added.

rschristian:
Haven't yet looked at the changes, but in a quick test, WMR's cache gets pretty broken if you were to alias to a parent directory (`"foo/": "../"`). Sub dirs (`"foo/": "../foo/"`) break all module resolution (no files can be found, even those in `public/`), and if you use an alias that points into the parent at all (`"foo/": "../"`) you lose all module reloading in the dev server. Have to restart on every change.

Not sure if this is in the scope of this PR, as that cache issue probably also ties into #589, but first thing I tried was aliasing to a parent directory, and I imagine some other users will as well.

marvinhagemeister:
Thanks for the review! Hadn't thought about nesting aliased paths, that's very valueable feedback! I'll look into it.

Regarding path aliasing in `tsconfig.json`: Maybe that's an indicator that we should just pick aliases up from that file and share the same syntax. ESBuild does something similar and it's kinda neat.

rschristian:
Oh cool, I didn't know ESBuild supported that.

Not sure about best UX here though, as we'd probably have to end up supporting both methods for users that don't want to use TS at all and that probably presents a few foot-guns in the implementation details.

An example path / support for `src/` in `create-wmr`'s template is probably a decent enough start.

marvinhagemeister:
@rschristian Alright, aliasing is now working for parent directories as well. You were right on the money that we wrote some invalid paths to the module graph and our own rollup shim.

Yeah we need to support both non-TS and TS users. I'm thinking we can just read the `tsconfig.json` file for any aliased paths and overwrite those if the same mapping name exists in one of our config files. It seems to cover most bases.

Open TODOs:
- [ ] Don't rewrite paths to aliases if it's not necessary

rschristian:
Another thing I just thought of, though maybe this is a better topic to bring up in slack as this is probably outside the scope of the PR here, is the goal only path aliases, or aliases in general? I can see some potential confusion coming up if we ship an `alias` section in `create-wmr`'s template and have a separate option called `aliases` in the config.

marvinhagemeister:
They should be the same. The config object supports both path aliases and package aliases. Seems like we have mixed wording there and we should unify them.

rschristian:
I did test that a couple days ago and while it worked (as far as I knew), I was seeing warnings being thrown. Not sure if that was resolved with your changes yesterday, will take a glance in the morning

rschristian (approved):
Fixes all the issues I mentioned earlier, looks great

marvinhagemeister:
Going to merge this. The remaining improvements are all issues or inconsistencies that were present before this PR and this PR is already longer than I'd like. Much easier to land smaller PRs against `main` 👍

Issue #254, opened 2020-12-15
CSS relative imports are not updated after build


rschristian:
Apologies if this is covered somewhere or already noted -- I couldn't find it if it was.

It seems that relative imports in CSS files are not currently handled in any way, which is an issue as some CSS files change directories. For example, I have a `/styles/global.css` mentioned in my `index.html`. This CSS file imports another CSS file via `@import 'normalize.css';`. This is fine in dev.

In prod, the issue is that while `global.css` gets moved into `dist/assets/styles/`, `normalize.css` is still sat in `dist/styles/normalize.css` and `global`'s import path doesn't change. It still tries to import `normalize` as if they were in the same directory.

This will effect CSS file chains like mentioned above, font file imports, images, etc. Perfectly possible I've set something up incorrectly somewhere too, though I'm using a vanilla WMR setup; I've only switched it to TS.

JoviDeCroock:
I guess this is due to the `optimize-graph` plugin opting out but us not moving it into the new folder when opting out, can look into it more later this week.

PR #558, opened 2021-04-27
Fix stale entry in WRITE_CACHE on file change

marvinhagemeister:


developit (approved):

developit:
I think there may even be an open bug report that this fixes? (maybe #254?)

PR #560, opened 2021-04-27
fix: don't forward to next after responding

ForsakenHarmony:


marvinhagemeister (approved):

packages/wmr/src/start.js
```diff
@@ -166,7 +166,6 @@ const injectWmrMiddleware = ({ cwd }) => {
 		} catch (e) {
 			next();
 		}
-		next();
```
marvinhagemeister, line 169: Wait on second thought: Do we still need to call `next()` in the success case when we don't end the request?

ForsakenHarmony: Oh right, we do, kinda missed that if 🙈

developit: I think this needs to be:

```js
try {
  if (...) {
    res.end(result);
    return;
  }
} catch (e) {}
next();
```

ForsakenHarmony: Oh, I moved out the if now

marvinhagemeister (approved):

PR #561, opened 2021-04-27
chore: make cli easier to read and add uncaught exception handler

ForsakenHarmony:


packages/wmr/src/cli.js
```diff
@@ -45,21 +52,25 @@ prog
 	.action(opts => {
 		opts.optimize = !/false|0/.test(opts.compress);
 		opts.compress = bool(opts.compress);
-		if (/true/.test(process.env.PROFILE)) opts.profile = true;
+		if (/true/.test(process.env.PROFILE || '')) opts.profile = true;
```
ForsakenHarmony, line 55: TS was complaining about this, but I can also revert it

packages/wmr/src/cli.js
```diff
@@ -45,21 +52,25 @@ prog
...
-			.map(frame => frame.raw)
-			.join('\n');
+	p.catch(catchException);
+}
 
-		process.stderr.write(`\n${kl.red(text)}\n${stack ? kl.dim(stack + '\n\n') : ''}`);
-		process.exit(p.code || 1);
-	});
+function catchException(err) {
+	const text = (err.stack && err.stack.split('\n')[0]) || err.message || err + '';
```
ForsakenHarmony, line 66: The first line of the stack had a nicer error than the actual message with the error I ran into

```
stack[0] = "Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client"
message  = "Cannot set headers after they are sent to the client"
```

marvinhagemeister (approved):

PR #562, opened 2021-04-27
Add support for loading env variables from `import.meta.env`

### LINKED ISSUE #457: Move `process.env` to `import.meta.env`
**Is your feature request related to a problem? Please describe.**

Move `process.env` to `import.meta.env`. It's more in line with the spirit of ES Modules and doesn't rely on us shimming a global variable.

**Describe the solution you'd like**
Store environment vars at `import.meta.env`. We'll probably need to keep transpiling `process.env` for a good while though as it's everywhere in the React ecosysstem.

**Describe alternatives you've considered**
-

**Additional context**

Both snowpack and vite store env variables at `import.meta.env`

marvinhagemeister:
Fixes #457

### DIFF
--- a/packages/wmr/src/plugins/process-global-plugin.js
+++ b/packages/wmr/src/plugins/process-global-plugin.js
@@ -1,3 +1,28 @@
+import { transform } from '../lib/acorn-traverse.js';
+
+/**
+ * Plugin to replace `process.env.MY_VAR` or `import.meta.env.MY_VAR` with
+ * the actual value.
+ * @param {Record<string, string>} env
+ */
+function acornEnvPlugin(env) {
+	return () => {
+		return {
+			name: 'transform-env',
+			visitor: {
+				MemberExpression(path) {
+					const source = path.getSource();
+					const match = source.match(/^(?:import\.meta|process)\.env\.(.+)/);
+
+					if (match) {
+						path.replaceWithString(JSON.stringify(env[match[1]]));
+					}
+				}
+			}
+		};
+	};
+}
+
 /**
  * Inject process globals and inline process.env.NODE_ENV.
  * @param {object} [options]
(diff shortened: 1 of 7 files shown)


packages/wmr/src/plugins/process-global-plugin.js
```diff
@@ -18,15 +43,13 @@ export default function processGlobalPlugin({ NODE_ENV = 'development', env = {}
...
-			code = code.replace(
-				/([(){}&|,;=!]\s*)process\.env\.NODE_ENV\s*([!=]==?)\s*(['"])(.*?)\3/g,
-				(str, before, comparator, quote, value) => {
-					let isMatch = value == NODE_ENV;
-					if (comparator[0] == '!') isMatch = !isMatch;
-					return before + isMatch;
-				}
-			);
+
+			const result = transform(code, {
```
developit, line 47: It looks like there's already a hook for injecting `import.meta.XX` properties we could use, which nicely avoids the transform step:

```js
	{
		name: 'process-global',
		// etc
		resolveImportMeta(property) {
			if (property === 'env') {
				return processObj;
			}
			return null;
		},
		transform(code) {
			// etc
		}
	};
```

Though this injects the whole process object anywhere `import.meta.env` is used, which isn't awesome. I wonder if we could combine the approaches?

```js
	return {
		name: 'process-global',
		// replace `import.meta.env.foo` with `process.env.foo`
		resolveImportMeta(property) {
			if (property === 'env') {
				return 'process.env';  // maybe '$importmeta.env'?
			}
			return null;
		},
		transform(code) {
			if (code.match(/[^\w$]process\.env/)) {
				code = `import process from '\0builtins:process.js';${code}`;
			}
		}
	};
```

(just thoughts fwiw, I'm not actually sure the transform is that expensive, since it's not doing any AST serialization 👍)

marvinhagemeister: Saw that hook, but it's not really useful in our situation as it only runs _after_ `transform()`. Because of that our `process` detection is too late and we'll never add that import.

marvinhagemeister:
FYI: Added a test case for production builds.

packages/wmr/src/plugins/process-global-plugin.js
marvinhagemeister: A bit unrelated: I'm actually hoping we'll switch to more ast-type transformations as they are harder to get wrong.

Rollup seems to allow plugins to return the `ast` directly [in a transform hook](https://rollupjs.org/guide/en/#transform). I wonder if this could be a mechanism to ensure that we'd only ever have one AST parse and serialization per file in the long term, despite multiple plugins participating in the transformation.

PR #569, opened 2021-05-01
Fix bare specifiers always rewritten as npm packages

marvinhagemeister:
This PR allows plugins to pick any name for a virtual module instead of having to dance around our npm or prefix detection logic. They can now just be named `my-virtual-module` or `@my-virtual-module` for example and it works out of the box. This is the main friction point I run into over and over whenever I play with custom plugins.

```js
{
  name: 'my-plugin',
  resolveId(id) {
    if (id === 'my-module') {
      return id;
    }
  },
  load(id) {
    if (id === 'my-module') {
      return `export const foo = 42`;
    }
  }
}
```

The detection is based on the return value of `load()`. In the case of virtual modules we'd return a non-null result, since plugins neat to load that. When the return value is `null` we therefore know that the passed `id` can't be a virtual module and therefore must refer to an npm package. Basically this is the same detection that rollup does internally.

There is one minor tradeoff to this approach though: Now any direct dependenants of a module are eagerly loaded. So if there is a module with 1000 conditional lazy imports we'd start loading all of those instead of only the one that the application requested. This happens because we can't know if a module is virtual or not before calling `.load(id)`.

An alternative to the implementation proposed here could be to skip trying to detect if a module is virtual or not until it is request. We currently do that detection logic when processing imports. So instead of rewriting paths to `/@npm/my-module`, we'd put them behind a generic prefix and resolve it to either a virtual module or npm package on load. It seems more elegant on paper and something we should definitely do in the long run, but I don't have a good grasp of the necessary changes required for that yet.

### DIFF
--- a/.changeset/sweet-tables-impress.md
+++ b/.changeset/sweet-tables-impress.md
@@ -1,0 +1,5 @@
+---
+'wmr': patch
+---
+
+Allow plugins to pick any specifier for virtual modules
--- a/packages/wmr/src/wmr-middleware.js
+++ b/packages/wmr/src/wmr-middleware.js
@@ -135,6 +135,16 @@ export default function wmrMiddleware(options) {
 			return next();
 		}
 
+		// Workaround for transform forcing extensionless ids to be
+		// non-js
+		let hasIdPrefix = false;
+
+		// Path for virtual modules that refer to an unprefixed id.
+		if (path.startsWith('/@id/')) {
+			hasIdPrefix = true;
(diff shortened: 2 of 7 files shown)


JoviDeCroock (approved):

developit (approved):
_Echoing from slack for posterity: this feels like the right current solution, but I am hoping that non-npm-bare-specifiers become "solved for free" once we switch the `/@npm` handling to a standard rollup plugin instead of custom middleware._

PR #570, opened 2021-05-01
Fix assets referenced in CSS not copied in prod builds

marvinhagemeister:
When an asset is referenced in a CSS file we didn't copy the files over to our output directory.

```css
@import "foo.css"; /* <- wasn't copied */

body {
  background: url('foo.jpg'); /* <- wasn't copied either */
}
```


packages/wmr/src/plugins/wmr/styles-plugin.js
```diff
@@ -130,6 +135,33 @@ export default function wmrStylesPlugin({ cwd, hot, fullPath } = {}) {
 				source = transformCss(source);
 			}
 
+			// Note: `plugin.generateBundle` is only called during prod builds for
+			// CSS files. So we need to guard the url replacement code.
+			if (production) {
+				source = await transformCssImports(source, idRelative, {
```
developit, line 141: reminder for myself: would be cool to implement `transformCssImports` within the `transformCss` pass so that it can use the Stylis AST.

marvinhagemeister: @developit the middlewares in `stylis` are [expected to be synchronous](https://github.com/thysultan/stylis.js/blob/e6843c373ebcbbfade25ebcc23f540ed8508da0a/src/Middleware.js#L17-L18), but our resolution mechanism is async. I'm not sure how we could avoid parsing the CSS twice with that restriction.

JoviDeCroock (approved):

packages/wmr/src/plugins/wmr/styles-plugin.js
developit: Ah good point. It would have to be two-pass, which would be less than ideal.

PR #576, opened 2021-05-02
Prerendering: Remove incomplete fake document

marvinhagemeister:
Prerendering of meta tags and other elements in the documents head previously worked by relying on a custom implementation of the global `document` object. We only did add a tiny subset of the methods that are available in the browser, but doing so breaks nearly all server detection code in third party libs. Checked various libraries and the most common snippet to do that is a variation of:

```js
// This was always false in wmr because we supply a fake document global
const isServer = typeof document === "undefined";
```

Because we did patch in a global document object, this value is always `false`, even during SSR.

Another problem with that approach in genral is that we'll always have a biased subset of the API that may work with our code, but is not guaranteed to work with third-party libraries. Examples of those are our very own [hoofd](https://github.com/JoviDeCroock/hoofd) and [react-helmet](https://github.com/nfl/react-helmet).

This is more complicated by the fact that to check if a meta tag is already set libraries use the `document.querySelector` API. And I feel like providing that bloats up WMR unnecessarily and only leads us down the rabbit hole of adding more and more DOM APIs.

Instead a much more reliable approach is to ditch the whole idea of supplying a fake document in the first place and instead rely on consumers supplying the head value.

TODO:

- [x] Update `document` usage in `docs` site
- [x] Check prerendering of `style()` (should be fine, but just to be sure)

### DIFF
--- a/packages/wmr/src/lib/prerender.js
+++ b/packages/wmr/src/lib/prerender.js
@@ -87,7 +87,7 @@ async function workerCode({ cwd, out, publicPath }) {
 		throw Error(`Unable to detect <script src="entry.js"> in your index.html.`);
 	}
 
-	/** @typedef {{ type: string, props: Record<string, string>, children?: string }} HeadElement */
+	/** @typedef {{ type: string, props: Record<string, string>, children?: string } | string | null} HeadElement */
 
 	/**
 	 * @type {{ lang: string, title: string, elements: Set<HeadElement>}}
@@ -102,32 +102,31 @@ async function workerCode({ cwd, out, publicPath }) {
 	// const App = m.default || m[Object.keys(m)[0]];
 
 	/**
-	 * @param {Set<HeadElement>} elements
-	 * @returns {Set<string>}
+	 * @param {HeadElement|HeadElement[]|Set<HeadElement>} element
+	 * @returns {string} html
 	 */
-	function serializeHead(elements) {
-		return new Set(
-			Array.from(elements).map(e => {
-				if (typeof e !== 'string') {
-					const type = e.type;
-					let s = `<${type}`;
-					s += Object.keys(e.props)
-						.sort()
-						// Filter out empty values
-						.filter(p => e.props[p] != null)
-						.map(p => (p == 'children' ? '' : ` ${p}="${enc(e.props[p])}"`))
-						.join('');
-					s += '>';
-					let kids = e.props.children || e.children;
-					if (!Array.isArray(kids)) kids = [kids];
-					if (!/link|meta|base/.test(type)) s += `${kids.join('')}</${type}>`;
-					return s;
-				}
-				return e;
-			})
-		);
+	function serializeElement(element) {
+		if (element == null) return '';
+		if (typeof element !== 'object') return String(element);
+		if (Array.isArray(element)) return element.map(serializeElement).join('');
+		const type = element.type;
+		let s = `<${type}`;
+		const props = element.props || {};
+		let children = element.children;
+		for (const prop of Object.keys(props).sort()) {
+			const value = props[prop];
+			// Filter out empty values:
+			if (value == null) continue;
+			if (prop === 'children' || prop === 'textContent') children = value;
+			else s += ` ${prop}="${enc(value)}"`;
+		}
+		s += '>';
+		if (!/link|meta|base/.test(type)) {
+			if (children) s += serializeElement(children);
+			s += `</${type}>`;
+		}
+		return s;
 	}
-
 	// We start by pre-rendering the homepage.
 	// Links discovered during pre-rendering get pushed into the list of routes.
 	const seen = new Set(['/']);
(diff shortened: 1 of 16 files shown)


packages/wmr/src/lib/prerender.js
```diff
@@ -146,19 +120,44 @@ async function workerCode({ cwd, out, publicPath }) {
 				routes.push({ url, _discoveredBy: route });
 			}
 		}
-		const body = (result && result.html) || result;
+
+		/**
+		 * @type {{ lang?: string, title?: string, links?: any[], metas?: any[]}}
```
developit, line 125: Couldn't this just use HTML? Right now there's no way to inject a script or style element.

marvinhagemeister: Good point, haven't thought about style or script tags. How do you feel about an API like this:

```ts
interface HtmlHead {
  lang: string;
  title: string;
  // placeholder name, not sure what else to call it
  elements: string[]; // meta-/script-/style-tags pre-serialized to a string
}
```

packages/wmr/src/lib/prerender.js
```diff
@@ -108,12 +87,47 @@ async function workerCode({ cwd, out, publicPath }) {
...
+					let kids = e.props.children || e.children;
+					if (!Array.isArray(kids)) kids = [kids];
+					if (!/link|meta|base/.test(type)) s += `${kids.join('')}</${type}>`;
+					return s;
+				}
+				return e;
+			})
+		);
+	}
+
```
developit, line 130: Adding support for recursion (1/2)

```suggestion
	/**
	 * @param {HeadElement|HeadElement[]|Set<HeadElement>} element
	 * @returns {string} html
	 */
	function serializeElement(element) {
		if (element == null) return '';
		if (typeof element !== 'object') return String(element);
		if (element instanceof Set) element = Array.from(element);
		if (Array.isArray(element)) return element.map(serializeElement).join('');
		const type = element.type;
		let s = `<${type}`;
		const props = Object.keys(e.props).sort();
		let children = element.children;
		for (const prop of props) {
			const value = e.props[prop];
			// Filter out empty values
			if (value == null) continue;
			if (prop === 'children' || prop === 'textContent') children = value;
			s += ` ${p}="${enc(e.props[p])}"`;
		}
		s += '>';
		if (!/link|meta|base/.test(type)) {
			if (children) s += serializeElement(children);
			s += `</${type}>`;
		}
		return s;
	}
```

packages/wmr/src/lib/prerender.js
```diff
@@ -155,19 +166,27 @@ async function workerCode({ cwd, out, publicPath }) {
...
+		// TODO: Use a proper HTML parser here. We should definitely not parse HTML
+		// with regex :S
 
 		// Inject HTML links at the end of <head> for any stylesheets injected during rendering of the page:
-		let headHtml = [
-			document.head[BEFORE],
-			...new Set(head.filter(c => c.rel && c.href).map(c => `<link rel="${enc(c.rel)}" href="${enc(c.href)}">`)),
-			document.head[AFTER]
-		].join('');
+		let headHtml = head.elements ? Array.from(serializeHead(head.elements)).join('') : '';
```
developit, line 184: Adding support for recursion (2/2)

```suggestion
		let headHtml = head.elements && serializeElement(head.elements);
```

packages/wmr/src/lib/prerender.js
```diff
@@ -108,12 +87,47 @@ async function workerCode({ cwd, out, publicPath }) {
 		throw Error(`Unable to detect <script src="entry.js"> in your index.html.`);
 	}
 
+	/** @typedef {{ type: string, props: Record<string, string>, children?: string }} HeadElement */
```
developit, line 90:
```suggestion
	/** @typedef {{ type: string, props: Record<string, string>, children?: string } | string | null} HeadElement */
```

developit (commented):
Support for recursive serialization and returning any iterable type as `render().head`.

packages/wmr/src/plugins/optimize-graph-plugin.js
```diff
@@ -9,7 +9,7 @@ import { hasDebugFlag } from '../lib/output-utils.js';
 const DEBUG = hasDebugFlag();
 
 const DEFAULT_STYLE_LOAD_FN = '$w_s$';
-const DEFAULT_STYLE_LOAD_IMPL = `function $w_s$(e,t){document.querySelector('link[rel=stylesheet][href="'+e+'"]')||((t=document.createElement("link")).rel="stylesheet",t.href=e,document.head.appendChild(t))}`;
+const DEFAULT_STYLE_LOAD_IMPL = `function $w_s$(e,t){if(typeof document==='undefined'){wmr.ssr.head.elements.add({type:'link',props:{rel:'stylesheet',href:e}});}else{document.querySelector('link[rel=stylesheet][href="'+e+'"]')||((t=document.createElement("link")).rel="stylesheet",t.href=e,document.head.appendChild(t))}}`;
```
developit, line 12: super nitpicky and tbh I'm not even sure which I prefer

```suggestion
const DEFAULT_STYLE_LOAD_IMPL = `function $w_s$(e,t){typeof document=='undefined'?wmr.ssr.head.elements.add({type:'link',props:{rel:'stylesheet',href:e}}):document.querySelector('link[rel=stylesheet][href="'+e+'"]')||((t=document.createElement("link")).rel="stylesheet",t.href=e,document.head.appendChild(t))}`;
```

PR #580, opened 2021-05-03
Add filesystem helpers

developit:
@rschristian's #578 gave me an idea: we have a few cases where soon we'll be able to move to some of the lovely Node 14+ `fs.*()` methods like `fs.rm(p,{recursive:true})`, but for now we need to support Node 12. Adding a little "fs helpers" lib will make this super easy to change in the future, and lets us use the methods now.

### DIFF
--- a/packages/wmr/src/build.js
+++ b/packages/wmr/src/build.js
@@ -1,6 +1,7 @@
 import * as kl from 'kolorist';
 import { promises as fs } from 'fs';
 import path from 'path';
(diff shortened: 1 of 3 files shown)


marvinhagemeister (approved):

developit:
(I'm assuming we don't bother with changesets for refactor/internal PRs, ya?)

marvinhagemeister:
yeah it's not user facing so we don't need changesets for those.

PR #582, opened 2021-05-03
create-wmr: template dep versions automatically kept up-to-date

rschristian:
So as it turns out, `pkg-install` does in fact write to the `package.json` file.... so long as you remember to scaffold your project before installing 🙃

`create-wmr` now installs the very latest versions of the deps and it will end up in the user's package file in the form of `"^x.x.x"`

### DIFF
--- a/packages/create-wmr/src/index.js
+++ b/packages/create-wmr/src/index.js
@@ -6,7 +6,7 @@ import sade from 'sade';
 import ora from 'ora';
 import kleur from 'kleur';
 import pkgInstall from 'pkg-install';
-const { getPackageManager, install } = pkgInstall;
+const { install } = pkgInstall;
 const { dim, bold, cyan, red } = kleur;
 
 sade('create-wmr [dir]', true)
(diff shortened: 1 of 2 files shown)


packages/create-wmr/src/index.js
```diff
@@ -52,13 +52,13 @@ sade('create-wmr [dir]', true)
...
+		spinner.start('scaffolding new project...');
+		await scaffold(ctx);
+		spinner.succeed('project created!');
+
 		const packageManager = /yarn/.test(process.env.npm_execpath) ? 'yarn' : 'npm';
-		// @ts-ignore-next
-		await getPackageManager({ prefer: packageManager }).catch(() => {
-			process.stderr.write(`\n${red(`${packageManager} cannot be found`)}\n`);
-			process.exit(1);
-		});
```
rschristian, line 60: I believe I wrote this because installing with yarn was originally a flag, like used in `preact-cli`. This is irrelevant though with the current set up. You can't run `npm init` / `yarn create` without having npm/yarn available. NPX comes with NPM, so it too isn't an potential issue.

marvinhagemeister (approved):
Sweeeet 🙌 🎉 Let's add a changeset and merge this 👍

rschristian:
Does this need a changeset? Shouldn''t actually change anything for users, just reduces maintenance burden.

I'm still not clear on the lines between needing a changeset vs not 😅

marvinhagemeister:
Good point, I don't really have a strong opinion on this 😬

rschristian:
I suppose it does alter user code but idk where the distinction is there

marvinhagemeister:
Yeah it's pretty blurry and I don't think there is a right or wrong approach here. It's more of an opinion which way we pick.

PR #583, opened 2021-05-03
Use Route components in template for potential TS users

marvinhagemeister:
Although our current template is JS-only, I'd expect a good portion (if not most) of our users to opt into TypeScript. We do leverage some trickery to make a global `path` prop work, but it's more consisten with other components to opt for  `Route` instead.

In most user's mind a JSX element is equivalent to a function call and in that model it's strange to see additional props besides ref/key.

### DIFF
--- a/.changeset/fast-lions-explain.md
+++ b/.changeset/fast-lions-explain.md
@@ -1,0 +1,5 @@
+---
+'create-wmr': patch
+---
+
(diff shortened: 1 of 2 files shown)


rschristian (approved):
Totally agree, and especially with the template shipping a tsconfig, I know my editor has some issues using routes as if they were route components

PR #584, opened 2021-05-03
Switch from subpackage to main imports for `preact-iso`

marvinhagemeister:
As discussed in slack. Only added a changeset for `create-wmr` as all other packages only have README changes or changes in their example app.

### DIFF
--- a/docs/public/content/docs/index.md
+++ b/docs/public/content/docs/index.md
@@ -100,7 +100,7 @@ render(<App />, document.body);
 
 ```diff
 -import { render } from 'preact';
-+import { hydrate } from 'preact-iso';
++import { hydrate, prerender as ssr } from 'preact-iso';
 
 function App() {
   return <h1>Hello World!</h1>;
@@ -110,8 +110,7 @@ function App() {
 +hydrate(<App />);
 
 +export async function prerender(data) {
-+  // we use dynamic import to prevent this from being loaded in the browser:
-+  return (await import('preact-iso')).prerender(<App {...data} />);
++  return ssr(<App {...data} />);
 +}
 ```
 
--- a/docs/public/content/docs/index.md
+++ b/docs/public/content/docs/index.md
@@ -100,7 +100,7 @@ render(<App />, document.body);
 
 ```diff
 -import { render } from 'preact';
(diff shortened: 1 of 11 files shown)


docs/public/content/docs/index.md
```diff
@@ -100,7 +100,7 @@ render(<App />, document.body);
 
 ```diff
 -import { render } from 'preact';
-+import hydrate from 'preact-iso/hydrate';
```
marvinhagemeister, line 103: Don't be fooled by this diff of a diff :D

developit: LOL wow that really throws GitHub for a loop

packages/directory-plugin/example/public/index.js
```diff
@@ -1,36 +1,36 @@
...
-    </LocationProvider>
-  );
+	return (
+		<LocationProvider>
+			<div class="app">
+				<nav>
+					<a href="/">Home</a>
+					<a href="/about">About</a>
+				</nav>
+				<ErrorBoundary>
```
developit, line 19: Might be good to remove `ErrorBoundary` here too - it's no longer needed since the Router handles suspensions itself now.

packages/create-wmr/tpl/public/index.js
```diff
@@ -27,6 +25,6 @@ export function App() {
 hydrate(<App />);
 
 export async function prerender(data) {
-	const { default: prerender } = await import('preact-iso/prerender');
-	return await prerender(<App {...data} />);
+	const { prerender: render } = await import('preact-iso');
```
developit, line 28: I think we can just have folks import `prerender` directly now, since it's already internally code-split:

```js
import { hydrate, prerender as ssr } from 'preact-iso';

export async function prerender(data) {
  return ssr(<App {...data} />);
}
```

(might also be a good time to alias it as `ssr`)

marvinhagemeister: done

packages/directory-plugin/example/public/index.js
marvinhagemeister: Let's do that separately as it's unrelated to moving imports around.

PR #585, opened 2021-05-03
Support returning prerender data

JoviDeCroock:
This add support for returning a `data` key from the prerender, this enables us to for instance use [prepass](https://github.com/preactjs/preact-ssr-prepass) or a manual `fetch` call to get some data needed for rendering.

### DIFF
--- a/packages/wmr/src/lib/prerender.js
+++ b/packages/wmr/src/lib/prerender.js
@@ -172,6 +172,12 @@ async function workerCode({ cwd, out, publicPath }) {
 			if (result.head) {
 				head = result.head;
 			}
+
+			if (result.data && typeof result.data === 'object') {
+				body += `<script type="isodata">${JSON.stringify(result.data)}</script>`;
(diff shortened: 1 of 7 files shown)


packages/wmr/src/lib/prerender.js
```diff
@@ -172,6 +172,12 @@ async function workerCode({ cwd, out, publicPath }) {
 			if (result.head) {
 				head = result.head;
 			}
+
+			if (result.data && typeof result.data === 'object') {
+				body += `<script type="isodata">${JSON.stringify(result.data)}</script>`;
```
marvinhagemeister, line 177: Let's add a simple test similar to the `prerender-external` fixture that just checks if the passed data string shows up in the rendered html 👍

marvinhagemeister (approved):

danielweck:
Follow up bug report (missing `preact-iso` NPM package update): https://github.com/preactjs/wmr/issues/592

PR #600, opened 2021-05-06
Fix directory plugin not resolving in prod builds

developit:
Rollup's `PluginContext.resolve()` does not resolve directories, whereas our dev-time version does (because it doesn't/can't check for existence). Using `path.resolve()` fixes this.

**Note: I have not yet tested that this PR works!**

### DIFF
--- a/packages/directory-plugin/src/index.js
+++ b/packages/directory-plugin/src/index.js
@@ -14,11 +14,11 @@ function directoryPlugin(options) {
 		async resolveId(id, importer) {
 			if (!id.startsWith('dir:')) return;
 
-			const resolved = await this.resolve(id.slice(4) + '\0', importer, { skipSelf: true });
+			const resolved = path.resolve(options.cwd || '.', importer, id.slice(4));
+			const stats = await fs.stat(resolved);
+			if (!stats.isDirectory()) throw Error(`Not a directory.`);
 
-			if (resolved) {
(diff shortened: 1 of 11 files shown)


packages/directory-plugin/src/index.js
```diff
@@ -14,11 +14,11 @@ function directoryPlugin(options) {
 		async resolveId(id, importer) {
 			if (!id.startsWith('dir:')) return;
 
-			const resolved = await this.resolve(id.slice(4) + '\0', importer, { skipSelf: true });
+			const resolved = path.resolve(options.cwd || '.', importer, id.slice(4));
```
marvinhagemeister, line 17: This might break any sort of aliasing users might do:

```
import whatever from "dir:~/foo";
```

To be able to process those paths we'd need to keep calling into `this.resolve()`

packages/directory-plugin/src/index.js
```diff
@@ -14,11 +14,11 @@ function directoryPlugin(options) {
...
+			const stats = await fs.stat(resolved);
+			if (!stats.isDirectory()) throw Error(`Not a directory.`);
```
marvinhagemeister, line 19: +1 for the directory check!

marvinhagemeister (commented):
Looking good! Left some comments and I think we should add a test case too. Let me know what you think

PR #603, opened 2021-05-06
Fix hoofd being inlined into main entry

marvinhagemeister:


JoviDeCroock (approved):

examples/demo/public/index.tsx
```diff
@@ -1,5 +1,4 @@
 import { LocationProvider, Router, lazy, ErrorBoundary, hydrate, prerender as ssr } from 'preact-iso';
-import { toStatic } from 'hoofd/preact';
```
danielweck, line 2: strictly speaking, 'prerender' (alias 'ssr') should also be dynamically imported to avoid "polluting" the browser runtime bundle. The doc site uses a separate 'prerender.js' file to isolate SSR, perhaps the demo site could promote this design pattern too?
https://github.com/preactjs/wmr/blob/1de2d7e543c4217467f3e9bf3e09449ec46131fc/docs/public/index.js#L33

marvinhagemeister: Good idea! I've opted for the approach from the docs site 👍

PR #606, opened 2021-05-07
Fix sass resolution error in production

marvinhagemeister:
Trying to narrow down a bug where sass imports fail

- [ ] Narrow down repro further
- [ ] Add test case
- [ ] Fix the bug!


marvinhagemeister:
well turned out to be an incorrect import specifier 🙃

PR #610, opened 2021-05-09
(wmr): Fix cli flags

rschristian:
Few issues this solves:
  * `yarn wmr serve --help` printed out two competing descriptions for the `--out` flag (1 global, 1 local)
  * `--public` isn't relevant for `serve`'s functionality
  *  `--out` for the dev server seems kind of odd and not intended? Basically `.cache/` will contain only the node modules (`@npm/`) while all the app stuff gets moved into whatever the `--out` dir is set to. Spreads the contents of `.cache/` over two directories which seems like an odd thing to want.

### DIFF
--- a/.changeset/silver-needles-grin.md
+++ b/.changeset/silver-needles-grin.md
@@ -1,0 +1,5 @@
+---
+'wmr': patch
+---
+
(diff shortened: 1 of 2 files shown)


marvinhagemeister (approved):
Good catch 👍

PR #611, opened 2021-05-10
chore: Update rollup-plugin-visualizer to latest

rschristian:
Upgrading will (eventually) allow us to exclude some portions of the bundle that aren't relevant to users. 

TODO:
- [x] Need a test for the visualizer going forward, as it relies on a pretty sketchy regex replace setup
  - Just that `wmr build --visualize` outputs a `stats.html` should be enough. We just need to make sure its assets are inlined properly at build time.
- [x] Need to bump `rollup` to >= 2.44 in order to return gzip and/or brotli stats to the user with `rollup-plugin-visualizer` >= 5


marvinhagemeister:
@rschristian hope you don't mind. Took the liberty of updating this PR.

rschristian:
@marvinhagemeister Oh, of course not. Has anyone ever minded not having to write tests themselves? 😄

I had put this on the back burner as I feel that the regex filtering (the feature gained by updating the visualizer) isn't particularly useful at the moment. A bit unintuitive and filtering doesn't result in the displayed stats updating. Might change in the future though.

Certainly could merge this in now though if you'd like? Nothing's holding it back.

marvinhagemeister:
I've found the root cause for why the tests were failing. It looks like the `external` option during resolution has changed in rollup `> 2.43.1`. See: https://github.com/rollup/rollup/pull/4021/

marvinhagemeister (approved):

PR #619, opened 2021-05-13
Deprecate `options.aliases` in favor of `options.alias`

marvinhagemeister:
As @rschristian noticed in #552 we sometimes check for `alias` and sometimes for `aliases`. We should only have one name for that feature. Just checked with other bundlers and the majorities use `alias`:

| Tool | Naming |
|---|---|
| [webpack](https://webpack.js.org/configuration/resolve/#resolvealias) | `alias` |
| [Parcel](https://parceljs.org/module_resolution.html#aliases) | `alias` |
| [vite](https://vitejs.dev/config/#resolve-alias) | `alias` |
| wmr | `alias` in `package.json`, otherwise `aliases` |

So to be consistent with everyone else (and ourselves), this PR deprecates `options.aliases` in favor of `options.alias`. Setting up aliasing via `options.aliases` will continue to work, but we'll print a deprecation warning to the console.

```bash
Deprecated: Found "aliases" property in WMR's configuration. It will be removed in a future version of WMR. Please switch to "alias" instead.
```

### DIFF
--- a/.changeset/light-flies-melt.md
+++ b/.changeset/light-flies-melt.md
@@ -1,0 +1,14 @@
+---
+'wmr': minor
+---
+
+Deprecate "aliases" configuration in favor of "alias" to have a consistent name across various frontend tools. The "aliases" property will continue to work, but a deprecation warning will be printed to the console.
+
+```diff
+export default {
+-	aliases: {
++	alias: {
+		react: 'preact/compat'
+	}
+};
+```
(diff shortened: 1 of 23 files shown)


rschristian (approved):
On a related note, how do people feel about deprecating setting aliases in the `package.json`? Moving all WMR configuration strictly into `wmr.config` sounds like a good move IMO.

Less moving parts to deal with, less potential for a user to accidentally omit important config information in a bug report, and I don't think having multiple ways to set the same value is particularly useful.

marvinhagemeister:
Agree, we should revisit supporting alias definitions in `package.json`. If my memory serves me correctly this was added long before wmr even had a configuration file.

PR #623, opened 2021-05-14
Fix/config typescript cjs

rschristian:
`compile-single-module` was hard-coded to return in ES format, which might not always be the case for an app


packages/wmr/src/lib/compile-single-module.js
```diff
@@ -27,55 +27,59 @@ let cache;
...
+			].filter(Boolean)
+		});
+		cache = bundle.cache;
+		const result = await bundle.write({
+			...ROLLUP_FAST_OUTPUT,
+			dir: out,
+			assetFileNames: '[name].[ext]',
+			paths: str => str,
+			format,
+			exports: 'auto'
```
rschristian, line 78: Rollup will spit out the following message if this is not set:

```
Entry module "wmr.config.ts" is implicitly using "default" export mode, which means for CommonJS output that its default export is assigned to "module.exports". For many tools, such CommonJS output will not be interchangeable with the original ES module. If this is intended, explicitly set "output.exports" to either "auto" or "default", otherwise you might want to consider changing the signature of "wmr.config.ts" to use named exports only.
```

`auto` should be fine for us, explicitly setting it silences the warning.

marvinhagemeister (approved):
💯

PR #624, opened 2021-05-14
Fix/typescript define config

rschristian:
`wmr` would get reassigned to `/_wmr.js` on anything passed through `compile-single-module` which would ultimately break `wmr.config.ts` files when using `defineConfig`


marvinhagemeister (approved):
Great catch 🙌

PR #625, opened 2021-05-14
Fix/config extension priority

rschristian:
If there's an issue in the transpiled output, say like in #624, the user will have a `wmr.config.js` in addition to their `wmr.config.ts`, as WMR encountered an error before it could clean that up.

With the existing priority, WMR would end up using that broken output before trying to use the `.ts` config. Users would have to manually remove the bad file before starting WMR again to ensure a clean run.

### DIFF
--- a/.changeset/shaggy-peaches-attend.md
+++ b/.changeset/shaggy-peaches-attend.md
@@ -1,0 +1,5 @@
+---
+'wmr': patch
+---
+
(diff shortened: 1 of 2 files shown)


JoviDeCroock (approved):

marvinhagemeister (approved):
This is great 🙌💯

PR #630, opened 2021-05-16
Add support for import assertions

marvinhagemeister:
Just played around with it, but there are some open issues to tackle:

1. `es-module-lexer` needs to support import assertions.
2. Our eslint setup needs to support import assertion (couldn't find a parser plugin)
3. How do we wire this into our plugin system?

Point 3. in particular is interesting because we may need to change behaviour of plugins depending on whether an import assertion is present or not. Take this example:

```js
import json from './foo.json' assert { type: 'json' };
```

Right now we transform the specifier to `/@json/foo.json` and serve a javascript file with the content

```js
export default { foo: "bar" }
```

This fails the assertion as the content type is now `text/javascript` instead of the expected one for json. So we likely want to only do the conversion to `/@json` when no import assertion is present. But this poses the question how we can detect that in a plugin? One possibility could be to extend the call signature of `resolveId`. Rollup has a third `options` parameter where we could add that:

```js
{
  name: "my-plugin",
  resolveId(id, importer, options) {
    if (id.endsWith(".json") && options.assert.type === "json") {
      // Assertion is present, do nothing
      return null;
    }

    // ... usual logic
  }
}
```


developit:
@marvinhagemeister based on your description, I'm assuming the goal here is to support passing through import assertions. I'm wondering if it's too early for that given this is only shipped in Chrome 91 - maybe for now we should transpile to our own json plugin internally?

marvinhagemeister:
@developit Good point. Agree with it being too early. I'll work on transpiling that syntax as you suggested 👍

marvinhagemeister:
Note: Found an issue when the the import assertion is inside an `async` iife. See https://github.com/xtuc/acorn-import-assertions/pull/1

marvinhagemeister:
Ok, only thing left is a problem during production builds. For some reason it errors with `unexpected token` due to a mismatch in references of an object. The shape is the same though. I have no clue why that is and debugging anything related to acorn is a nightmare.

marvinhagemeister:
Traced the remaining node 12 + rollup build issue down to needing to support properly support esm in the acorn plugin. Upstream PR: https://github.com/xtuc/acorn-import-assertions/pull/5

PR #622, opened 2021-05-13
Fix missing whitespace when JSXText spans multiple lines

marvinhagemeister:
Fixes #621.

PR #633, opened 2021-05-17
Adapt the newline regex rule to match any char range

### LINKED ISSUE #632: Missing whitespace in multiline jsx-text
Is related to #621 
the bug is still here,
now with cyrillic text

After building this jsx, I see text without spaces
input:
```jsx
<p>this
is
sparta</p>
<p>это
домашняя
страница</p>
```
output:
```jsx
<p>this is sparta</p><p>этодомашняястраница</p>
```

the problem was detected only when I use the Cyrillic alphabet (with the Latin alphabet ok), and only when building (in dev-mode everything is ok)

cristianbote:
Fixes #632 and builds on top of #622

### DIFF
--- a/examples/demo/wmr.config.js
+++ b/examples/demo/wmr.config.js
@@ -1,0 +1,7 @@
+function wmr_config () {
+	return {
+		alias: {
+			'src/*': 'src'
+		}
+	};
+}export default wmr_config;
--- a/examples/demo/wmr.config.js
+++ b/examples/demo/wmr.config.js
@@ -1,7 +1,0 @@
-function wmr_config () {
-	return {
-		alias: {
-			'src/*': 'src'
-		}
-	};
-}export default wmr_config;
--- a/packages/wmr/src/lib/acorn-traverse.js
+++ b/packages/wmr/src/lib/acorn-traverse.js
@@ -489,7 +489,7 @@ const TYPES = {
 							// We can drop the whole matched string if it only contains
 							// whitespace characters, because that means we're at the
 							// beginning or at the end of the JSXText node
-							const replacement = /\w/.test(value) ? ' ' : '';
+							const replacement = /[.\S]/.test(value) ? ' ' : '';
 							value = value.replace(/\s*\n+\s*/g, replacement);
 						}
 						children.push(TYPES.stringLiteral(value));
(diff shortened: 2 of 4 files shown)


packages/wmr/src/lib/acorn-traverse.js
```diff
@@ -489,7 +489,7 @@ const TYPES = {
 							// We can drop the whole matched string if it only contains
 							// whitespace characters, because that means we're at the
 							// beginning or at the end of the JSXText node
-							const replacement = /\w/.test(value) ? ' ' : '';
+							const replacement = /[.\S]/.test(value) ? ' ' : '';
```
marvinhagemeister, line 492: Good catch about inverting the rege 👍 Quick question regarding including `.`: Did you found a case where this signals a newline?

cristianbote: Nah, this was all you mate!

Regarding `.` matching newlines, the dot notation says: "any character _except_ newline". So, that means also whitespace, and that's why I added the `\S`(not whitespace) in the range so we only get matching any chars expect: newlines and whitespaces.
Did I get it right?

marvinhagemeister: > Regarding `.` matching newlines, the dot notation says: "any character _except_ newline".

Shouldn't that be `[^\S]` instead of `[.\S]`?

cristianbote: Oh, I understand the confusion now. `\S` capital 'S' means _not_ whitespace, lowercase 's' means whitespace.

developit: `[.]` matches a literal `.` character (this one always throws me haha).

I think `[^\s]` works here:

```js
/[^\s]/.test(' ') // false
/[^\s]/.test('\n') // false
/[^\s]/.test(' \n \t') // false
/[^\s]/.test('a') // true
/[^\s]/.test('\na ')
```

cristianbote: Uh oh! 😲  What in the name of holly regex rules... TIL. Then yup! That's what it should be. Changing it.

examples/demo/wmr.config.js
```diff
@@ -0,0 +1,7 @@
+function wmr_config () {
+	return {
+		alias: {
+			'src/*': 'src'
+		}
+	};
+}export default wmr_config;
```
marvinhagemeister, line 7: This file doesn't look like it belongs to this PR.

marvinhagemeister (commented):
Looking good! One minor nit with the accidental commit of the transpiled config file and other than that we're good to go 👍

examples/demo/wmr.config.js
cristianbote: Thank you 🙏

marvinhagemeister (approved):

PR #635, opened 2021-05-18
Fix relative import specifiers becoming bare in development

marvinhagemeister:
This PR ensures that we keep passing relative import specifiers to resolveId. Previously we'd loose the leading `./`, leading to `./foo.js` being passed as `foo.js` during development.

This is a necessary precursor step towards turning the npm middleware into a standard plugin.

### DIFF
--- a/packages/wmr/src/wmr-middleware.js
+++ b/packages/wmr/src/wmr-middleware.js
@@ -183,6 +183,12 @@ export default function wmrMiddleware(options) {
 				path = prefixMatches[2];
 			}
 
+			if (path.startsWith('/@id/')) {
+				// Virtual paths have no exact file match, so we don't set `file`
+				hasIdPrefix = true;
+				path = path.slice('/@id'.length);
+			}
+
 			// convert to OS path:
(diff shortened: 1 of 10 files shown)


packages/wmr/src/wmr-middleware.js
```diff
@@ -490,7 +506,7 @@ export const TRANSFORMS = {
 						}
 					}
 
-					const modSpec = spec.startsWith('../') ? spec.replace(/..\/g/, '') : spec.replace('./', '');
```
marvinhagemeister, line 493: I have no clue how the `/..\/g/` regex ever made it into our code base. Its intention is pretty obvious, safe for fact that the regex syntax doesn't work that way in JavaScript :D

```txt
. matches any character (except for line terminators)
. matches any character (except for line terminators)
\/ matches the character / literally (case sensitive)
g matches the character g literally (case sensitive)
```

So this regex only matched literally strings like: `"fa/g"`

developit: heh - I should have caught that! looks like it snuck in in #552 🥸

developit (approved):

PR #636, opened 2021-05-21
(wmr): Fix debug in optimize-graph when debug is set via flag

rschristian:
`hasDebugFlag()` partially relies on `debugCliArg`, which can only be truthy after `normalize-options` is ran. `DEBUG` being assigned at the module level here means it gets its value before `normalize` and sticks with it even after.

Simply pushed the assignment back a bit


marvinhagemeister (approved):
Good catch 👍

PR #641, opened 2021-05-24
Fix CSS Modules mappings containing invalid identifiers

### LINKED ISSUE #640: css modules don't handle any special class names in development
**Describe the bug**
.js files generated for module'd css can't handle any classes that arent valid js variable names (`e.g. `.place-1` or `new`)

**To Reproduce**

1. Create file.module.css and import it
2. Set its content to `.new {color: #333} .test-z {color: red}`
3. Run dev wmr. It crashes with `new is not a valid token`

**Expected behavior**
Tokens need to be escaped, so that it doesn't fail.

developit:
Fixes #640

### DIFF
--- a/packages/wmr/src/plugins/wmr/styles-plugin.js
+++ b/packages/wmr/src/plugins/wmr/styles-plugin.js
@@ -185,7 +185,7 @@ export default function wmrStylesPlugin({ cwd, hot, fullPath, production, alias 
 				.map(m => {
 					const matches = m.match(/^(['"]?)([^:'"]+?)\1:(.+)$/);
 					if (!matches) return;
-					if (RESERVED_WORDS.test(matches[2])) return;
+					if (RESERVED_WORDS.test(matches[2])) return console.warn('Cannot use reserved word "%s" as classname; found in "%s"', matches[2], idRelative);
 					let name = matches[2].replace(/-+([a-z0-9])/gi, (s, c) => c.toUpperCase());
 					if (name.match(/^\d/)) name = '$' + name;
 					return name + '=' + matches[3];
--- a/packages/wmr/src/plugins/wmr/styles-plugin.js
+++ b/packages/wmr/src/plugins/wmr/styles-plugin.js
@@ -4,6 +4,9 @@ import { transformCssImports } from '../../lib/transform-css-imports.js';
 import { transformCss } from '../../lib/transform-css.js';
 import { matchAlias } from '../../lib/aliasing.js';
 
+// invalid object keys
+const RESERVED_WORDS = /^(abstract|async|boolean|break|byte|case|catch|char|class|const|continue|debugger|default|delete|do|double|else|enum|export|extends|false|final|finally|float|for|function|goto|if|implements|import|in|instanceof|int|interface|let|long|native|new|null|package|private|protected|public|return|short|static|super|switch|synchronized|this|throw|throws|transient|true|try|typeof|var|void|volatile|while|with|yield)$/;
+
 /**
  * @param {string} sass
  * @returns {string} css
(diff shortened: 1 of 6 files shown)


packages/wmr/src/plugins/wmr/styles-plugin.js
```diff
@@ -4,6 +4,9 @@ import { transformCssImports } from '../../lib/transform-css-imports.js';
 import { transformCss } from '../../lib/transform-css.js';
 import { matchAlias } from '../../lib/aliasing.js';
 
+// invalid object keys
+const RESERVED_WORDS = /^(abstract|async|boolean|break|byte|case|catch|char|class|const|continue|debugger|default|delete|do|double|else|enum|export|extends|false|final|finally|float|for|function|goto|if|implements|import|in|instanceof|int|interface|let|long|native|new|null|package|private|protected|public|return|short|static|super|switch|synchronized|this|throw|throws|transient|true|try|typeof|var|void|volatile|while|with|yield)$/;
```
lukeed, line 8: makes you wish this list were available somewhere as a global :D

lukeed (approved):

Inviz:
Honestly i'm not sure what's going on here with reserved names. I think it should escape them, so `style.new` would work, but it instead skips them? If it was me, i'd output class names as they are escaped `"new": "...", "new-0": "..."`, and additionally transformed like `New: "", new0: ""` as well for named import. Skipping them does not seem like a good idea

lukeed:
Using reserved JS keywords as classnames (and keynames, by extension) is already disallowed in a lot of CSS setups; eg [`css-loader`](https://github.com/webpack-contrib/css-loader#namedexport)

Inviz:
Well silently skipping them is just sadistic to users. Personally I am not importing them by name, and use `style.***` syntax, and sometimes like `style['place-${i}']` which is pretty valid.

packages/wmr/src/plugins/wmr/styles-plugin.js
```diff
@@ -182,7 +185,8 @@ export default function wmrStylesPlugin({ cwd, hot, fullPath, production, alias
 				.map(m => {
 					const matches = m.match(/^(['"]?)([^:'"]+?)\1:(.+)$/);
 					if (!matches) return;
-					let name = matches[2].replace(/-+([a-z])/gi, (s, c) => c.toUpperCase());
+					if (RESERVED_WORDS.test(matches[2])) return;
```
lukeed, line 188:
```suggestion
					if (RESERVED_WORDS.test(matches[2])) return console.warn('Cannot use reserved word "%s" as classname; found in "%s"', matches[2], idRelative);
```

Inviz: this makes more sense, thanks.

marvinhagemeister (commented):
Looks good, but let's add a test case to ensure that we don't regress on this in the future 👍

developit:
Ah - so to clarify @Inviz, this only skips the named exports, the default-exported styles object is still classNames verbatim from source.

Personally I think it might be better if we just removed named exports of classes. Would anyone disagree?

Inviz:
> Ah - so to clarify @Inviz, this only skips the named exports, the default-exported styles object is still classNames verbatim from source.
I see, it makes sense. Keep stuff as is in the object.
> 
> Personally I think it might be better if we just removed named exports of classes. Would anyone disagree?
I dont use named imports, so I would not mind this at all.

marvinhagemeister:
FWIW other bundlers seem to just silently drop exposing const exports for reserved variable names.

marvinhagemeister (approved):

Issue #645, opened 2021-05-25
processSass actually doesn't process sass


Inviz:
**Describe the bug**
TRANSFORMS.css invokes `processSass` but that never transforms the sass. That leads to non-modular sass files being served as-is, and break on HMR too. They *may* work only if included in `.js` file, not using aliases and until first HMR.

**To Reproduce**

Alias issue:
1. Create new wmr project, install sass/node-sass
2. Rename style.css to style.scss
3. Use some sass variables: `$color: red; body { color: $color} ` 
4. Add alias `"public/*": "public"`
5. Run wmr, color is not red

HMR non-module issue:
1. Create new wmr project, install sass/node-sass
2. Rename style.css to style.scss
3. Use some sass variables: `$color: red; body { color: $color} ` 
4. Change reference in index.html to `style.scss`
5. Variable doesn't work: `style.scss` contains raw text `$color:red;body{color:$color !important;}`
6. Remove style.scss from index.html 
7. Add `import ./style.scss` to index.js
8. Restart wmr
9. Variable works! `body{color:red!important;}` 
10. Re-save style.scss to trigger HMR
11. Variable doesn't work: `style.scss` contains raw text `$color:red;body{color:$color !important;}`
12. Hard refresh page
11. Variable doesn't work: `style.scss` contains raw text `$color:red;body{color:$color !important;}`

**Expected behavior**
Variables should always work, sass should run on all sass file changes

**Desktop (please complete the following information):**

- OS: [ MacOS Big Sur]
- Browser: [ chrome]
- Node Version: [16]
- WMR Version: Latest

Inviz:
I started a PR that adds a demo that displays the issues. In the demo it was not even possible to reproduce it working at all, inside lazy() or something it does not work period.

Inviz:
OK i made a hack that works for my case, running sass plugin explicitly in TRANSFORMS.css. It feels like this css support needs to be revised quite a bit.

ettoredn:
fun fact: they literally deleted my issue stating the same.

marvinhagemeister:
> fun fact: they literally deleted my issue stating the same. Way to go.

Please stop spreading nonsense, that's not welcome here. The only issue you ever opened on this repo is #670. So let's stay on topic and work on getting this resolved.

PR #646, opened 2021-05-25
Add demo for sass support / Fix non-module sass file support HMR

Inviz:
This adds a demo page for sanity check of sass support. This does not have a fix for https://github.com/preactjs/wmr/issues/645, but it will help checking it if somebody figures it out.

### DIFF
--- a/examples/demo/package.json
+++ b/examples/demo/package.json
@@ -17,14 +17,15 @@
 		"react": "preact/compat"
 	},
 	"devDependencies": {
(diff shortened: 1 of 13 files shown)


Inviz:
So by the way, the demo app doesn't have sass installed or bundled, amirite? I think this means that the sass test that is in there at the moment does nothing

developit:
@Inviz correct. I think we might want to move the sass demo into a test fixture, which have their own deps.

PR #648, opened 2021-05-27
fix(polkompress): avoid hanging requests and/or runtime errors

lukeed:
Ok, this one is a little long-winded:

The `polkompress` middleware didn't always "cleanup"/terminate the response correctly. Basically, our monkey-patched `res.writeHead` held a dangling `pendingStatus` value for when `start()` was done. However, this only worked if there was an _explicit_ call to `res.writeHead` somewhere in the application _before_ the `start()` exits.

I don't know if this just has to do with the fact that my test case used a small `body`  (even when crossing the threshold) , but in my reproductions I saw that this was always the order:

```js
INSIDE END // via res.end patch
START END { pendingStatus: undefined } // last line of start()
INSIDE WRITEHEAD 200 // via res.writeHead patch
```

So, the `res.writeHead` call that `res.end` makes internally (native, not patched behavior) happens after `start()` has already finished. This problem manifested as two different errors:

1) When body was shorter than the threshold, it results in a hanging request that never terminates – at least until timeout takes over

2) When the body crosses the threshold (and _does not_ take a while to finish compressing, I guess), this nasty error surfaces, which was seen in production (repro below):

Screen_Shot_2021-05-26_at_8 27 29_PM

You can get to both errors with this snippet & by playing with the `res.end` contents:

```js
http.createServer((req, res) => {
  compression({ level: 1, threshold: 4 })(req, res, () => {
    res.end("OK");
    // res.end("HELLO WORLD");
  })
}).listen(3000);
```

I added tests (nearly crashed my computer, twice!!) for both variants of this error.

### DIFF
--- a/.changeset/hungry-students-pump.md
+++ b/.changeset/hungry-students-pump.md
@@ -1,0 +1,5 @@
+---
+'wmr': patch
+---
+
+fix(compression): ensure all responses write headers
--- a/packages/wmr/src/lib/polkompress.js
+++ b/packages/wmr/src/lib/polkompress.js
@@ -67,7 +67,7 @@ export default function compression({ threshold = 1024, level = -1, brotli = fal
 				listeners.forEach(p => on.apply(res, p));
 			}
 
-			if (pendingStatus) writeHead.call(res, pendingStatus);
+			writeHead.call(res, pendingStatus || res.statusCode);
 		}
 
 		const { end, write, on, writeHead } = res;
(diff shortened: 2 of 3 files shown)


marvinhagemeister (approved):
What a great catch! Can only imagine how long it took to debug this 👍

maraisr (approved):
🚀👽 to the moon

developit:
Wow, awesome debugging here.

PR #649, opened 2021-05-28
Do not inject link tag if entry css not exists

leader22:
If `/public/index.html` does not have entry CSS file like this..

```html
<!DOCTYPE html>
<html lang="en">
	<head>
		<meta charset="utf-8">
		<title>repro</title>
	</head>
	<body>
		<script type="module" src="/index.js"></script>
	</body>
</html>
```

w/ minimal `/public/index.js`.

```js
import { hydrate, prerender as ssr } from 'preact-iso';

const App = () => <h1>Hello World!</h1>;

hydrate(<App />);

export const prerender = async (data) => ssr(<App {...data} />);
```

`wmr build --prerender` generates this HTML.

```html
<!DOCTYPE html>
<html lang="en">
	<head>
		<meta charset="utf-8">
		<title>repro</title>

		<link rel="stylesheet" href="/undefined"> <!-- 👀 -->
</head>
	<body><h1>Hello World!</h1>
		<script type="module" src="/index.68215b6e.js"></script>
	</body>
</html>
```

This empty `link` element should not be injected. Does this change make sense?

### DIFF
--- a/.changeset/sour-keys-attend.md
+++ b/.changeset/sour-keys-attend.md
@@ -1,0 +1,5 @@
+---
+"wmr": patch
+---
+
+Do not inject `<link rel=stylesheet>` when there is no stylesheet.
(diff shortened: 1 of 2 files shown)


developit (approved):
Good catch!

JoviDeCroock (approved):

leader22:
Thank you for the review+merge! 🙌🏻 

FYI: @cristianbote I just happened to notice this issue while browsing https://goober.rocks . When the next version is released, it will reduce the HTTP errors on your site. 😉

PR #651, opened 2021-05-30
Fix prerender deletes <html> attributes

### LINKED ISSUE #593: WMR 2.0.0 new `prerender()` returned `head` object with `lang` property caused all existing attributes to be removed (as documented in source code)
I know that you know :D

https://github.com/preactjs/wmr/blob/00ed4e1d428177b77fe5b98abc1e48ed739cb3e0/packages/wmr/src/lib/prerender.js#L203-L207

...but I just wanted to file an issue so that this problem isn't "ignored" due to other development priorities.

My workaround for this bug is to not set `lang` in `head`, and to supply the attribute directly in `index.html` (alongside the other attributes I need, notably `dir="ltr"`). There are of course other possible workarounds, such as setting `dir="ltr"` on `body` instead, but I have other attributes on the `html` root element which are used to enforce CSS specificity (notably: `id`).

So for me this is not a deal breaker as I can just set all required attributes statically in `index.html`, but this could be problematic for other devs.

marvinhagemeister:
Fixes #593

### DIFF
--- a/packages/wmr/src/lib/prerender-worker.js
+++ b/packages/wmr/src/lib/prerender-worker.js
@@ -1,0 +1,196 @@
+import { parentPort, workerData } from 'worker_threads';
+import { promises as fs } from 'fs';
+import path from 'path';
+
+/**
+ * @param {string} str
+ * @returns {string}
+ */
+function enc(str) {
+	return str.replace(/&/g, '&amp;').replace(/"/g, '&quot;').replace(/</g, '&lt;').replace(/>/g, '&gt;');
+}
+
+/** @typedef {{ type: string, props: Record<string, string>, children?: string } | string | null} HeadElement */
+
+/**
+ * @typedef {{ lang: string, title: string, elements: Set<HeadElement>}} HeadResult
+ */
+
+/**
+ * @param {HeadElement|HeadElement[]} element
+ * @returns {string} html
+ */
+function serializeElement(element) {
+	if (element == null) return '';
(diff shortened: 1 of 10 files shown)


packages/wmr/src/lib/prerender.js
```diff
@@ -44,178 +42,3 @@ export function prerender({ cwd = '.', out = '.cache', publicPath }) {
 		w.once('exit', resolve);
 	});
 }
-
-async function workerCode({ cwd, out, publicPath }) {
```
marvinhagemeister, line 48: The diff looks a lot larger due to moving this function into a separate file.

packages/wmr/src/lib/prerender-worker.js
```diff
@@ -0,0 +1,219 @@
...
+								hasTitle = true;
+								headNode.content = [head.title];
+							}
+							return headNode;
+						});
+
+						if (!hasTitle) {
+							// TODO: TS types of posthtml seem to be wrong
+							// @ts-ignore
+							node.content?.unshift({
```
developit, line 181: breaks in Node 12

```suggestion
							node.content && node.content.unshift({
```

danielweck:
Hello :)
I see that  @developit added the comment "breaks in Node 12"  on 30 May 2021. Is that the only blocker for this PR to be merged?
Issue #593 currently prevents me from using Hoofd ( https://github.com/JoviDeCroock/hoofd ) `useLang()` :(
Thank you all! 👍

PR #652, opened 2021-05-30
Terser's typings are (technically) correct

fabiosantoscode:
There's an undocumented option, `options.format.code`, which you can set to false to have Terser not output any code. In that case `out.code` is indeed undefined. It can be used if you're interested in the ESTree AST only (it's helpful if you want to do further processing to the code).

Side note, the Rust port of Terser built by the SWC project (which I've been neglecting a bit but I try to help out) is approaching a ready status. If you'd like to access an experimental version, let me know.

Much love!

### DIFF
--- a/packages/wmr/src/plugins/fast-minify.js
+++ b/packages/wmr/src/plugins/fast-minify.js
@@ -33,7 +33,6 @@ export default function fastMinifyPlugin({ sourcemap = false, warnThreshold = 50
 				return this.error(err);
 			}
 
-			// TODO: Check if tersers typings are wrong
 			if (!out.code) out.code = code;
 
 			if (duration > warnThreshold && hasDebugFlag()) {


developit (approved):

developit:
Heh - thanks for the sanity check @fabiosantoscode :D

I'd be interested in checking out the Rust port. TBH I don't find Terser slow in WMR currently, but in larger projects I believe it's the thing that takes the longest during production builds.

fabiosantoscode:
It does take a while :'( lots of RAM being used too. The thing with rust is that it enables better maintainability, due to the type system and mutability guards.

Also the fact that it's so fast lets SWC very nicely separate different transformers. Crawling the AST is no longer a bottleneck so it can afford to make the code more organized with little cost.

developit:
@fabiosantoscode any idea if the Rust port is usable without the rest of SWC? SWC is certainly compelling, just having a direct drop-in replacement for Terser would be most useful.

fabiosantoscode:
I'd love to create a drop-in replacement ASAP and I think it's very doable even without working from within SWC.

ForsakenHarmony:
> There's an undocumented option, `options.format.code`, which you can set to false to have Terser not output any code. In that case `out.code` is indeed undefined.

Wonder how hard that would be to express in the typings 🤔 (not saying you should do it at all, most likely not worth the effort, but with all I've seen it's probably possible)

PR #657, opened 2021-06-02
Fix sass files not being processed

marvinhagemeister:
Fixes the main issue mentioned in #645

### DIFF
--- a/.changeset/nice-clouds-carry.md
+++ b/.changeset/nice-clouds-carry.md
@@ -1,0 +1,5 @@
+---
+'wmr': patch
+---
+
+Fix `.sass` and `.scss` files not being processed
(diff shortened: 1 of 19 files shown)


marvinhagemeister:
Alright, imo this is ready to be merged.

Inviz:
Great stuff! I think it would be great to test nested imports from sass files (that define variables) and both hmr/build paths for completeness. That's so we can put this sass guy to rest

marvinhagemeister:
@Inviz Agree, the current test suite isn't extensive enough and needs to be enhanced to check for those cases 👍

marvinhagemeister:
Quick update: I think I have now a good overview of the necessary changes to be made to WMR to get this fixed. Whilst this PR fixes the compilation in itself, it doesn't fix HMR issues. For that we need to rework the CSS handling a bit and wire it up into the same plugin-style architecture as we do with the `.js` code.

Inviz:
Thanks Marvin. Yes, that's what I gathered myself, is that it seems that the easy *and* proper fix is out of reach for me and there needs to be a better underlying change for handling css. I think this is a right thing to do and it is very much appreciated

PR #661, opened 2021-06-02
chore: Init'd create-wmr app is private by default

rschristian:
Probably fair to mark newly created wmr apps as being private, not for publish on NPM

Yarn will alert the user to this on every single command, which tends to get mildly annoying after a certain point


marvinhagemeister (approved):
That makes a lot of sense. WMR is primary used with applications and those are rarely published to npm. Good catch 👍

PR #662, opened 2021-06-03
run-fixture: install dependencies before starting

marvinhagemeister:
This makes it easier to locally debug fixtures which use custom dependencies


developit:
Do we have any fixtures that are being used to test automatic dependency installation?

marvinhagemeister:
@developit with #657 which brings back support for `.sass/.scss` files we will have one. By default `sass` isn't installed so I've added a fixture which does that to be able to check if sass compilation actually works.

marvinhagemeister:
Closing, much easier if we just add it as a `devDependency`. See #657

PR #664, opened 2021-06-03
chore: Remove unused 'useEffect' import in preact-iso/router

rschristian:
Besides simply being unused, it causes this message in any new app / any app that uses `preact-iso/router`, which isn't really helpful.

temp


developit (approved):

rschristian:
Closing as I'm gonna roll this change into another PR

PR #665, opened 2021-06-04
Refactor/shared tooling

rschristian:
Stems from #664, an issue that ESLint would've caught had it been set up.

- Removes ESLint, Husky, and Prettier config from `packages/wmr`. It's all duplicated at the package root. We don't need it.
- Updates workflows
- Fixes low-hanging linting issues in `preact-iso`
- Includes `.ts` files in pre-commit hook so no more needing to run `prettier` manually on them

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,9 +83,6 @@
 			"pre-commit": "lint-staged"
 		}
 	},
-	"jest": {
-		"preset": "jest-puppeteer"
-	},
 	"devDependencies": {
 		"@changesets/changelog-github": "^0.2.7",
 		"@changesets/cli": "^2.12.0",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -68,7 +67,7 @@
 		"arrowParens": "avoid"
 	},
 	"lint-staged": {
-		"packages/**/{src,test}/**/*.js": [
+		"packages/**/*.{js,ts}": [
```
rschristian, line 70: I did have to run `npm rebuild` / `npm rebuild --update-binary` in order to make Husky play nice. Lint staged kept strangely erroring out otherwise.

marvinhagemeister (approved):

PR #666, opened 2021-06-04
ci: Adding yarn cache

rschristian:
Hopefully will speed up build times by a noticeable percentage. [This job took 50s to install](https://github.com/preactjs/wmr/runs/2742146763?check_suite_focus=true) out of 7 minutes total run time, for example.


rschristian:
Hm doesn't seem to be noticeably better.

PR #672, opened 2021-06-07
(docs) URLs with trailing slashes

developit:
It looks like Netlify automatically redirects to URLs with trailing slashes (go to `wmr.dev/docs`, then reload). I'm not sure if there's a way to turn that off on the Netlify side? If not, we can just generate them with trailing slashes to match.

### DIFF
--- a/docs/public/components/doc-structure.js
+++ b/docs/public/components/doc-structure.js
@@ -4,7 +4,7 @@ export { content };
 const pages = new Map(
 	content.map((item, index) => {
 		if ('heading' in item) return ['', -1];
-		item.slug = `/docs/${item.name}`.replace(/\/index$/g, '');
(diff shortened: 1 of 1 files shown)


marvinhagemeister (approved):

marvinhagemeister (commented):
On second thought: Maybe the trailing url thing is something we should deal with in our router instead? Most other ones seem to do it and that's a much more generic solution rather than tying the url structure to a specific hosting provider.

developit:
What do you mean by dealing with it in the router?

marvinhagemeister:
Nah, rather that the regex is more loose and matches both _with_ and _without_ trailing slashes for the same route.

```
/blog/:id -> /blog/post-a  = match
          -> /blog/post-a/ = match
```

(Might be misunderstanding the issue at hand when it comes to netlify)

PR #678, opened 2021-06-08
Extract compilation code from styles-plugin

marvinhagemeister:
Mainly reduces some noise in the styles plugin.


developit (approved):
This is a nice improvement.

PR #680, opened 2021-06-10
Fix swapped usage of `options.cwd` vs `options.root`

### LINKED ISSUE #142: Add `--cwd` + `--base` cli options
Following our conversation from today from slack:

- `--base` -> web root, default is `/public`
- `--cwd` -> the root directory to work in, default is `process.cwd()`

Lots of current code uses `cwd` for everything. We need to revisit which places need `base` and which `cwd.

marvinhagemeister:
This PR corrects the swapped usage of `options.cwd` vs `options.root`. The former is now always equal to the current working directory and `options.root` points to the absolute public dir (=usually `<cwd>/public`).

Marking as major, because users might have worked around the swapped usage in custom plugins and doing this in a non-major release would break their config.

For the directory plugin a major release is not necessary, because npm treats minor versions below `< 1.0.0` as major and won't update to that automatically.

Fixes #142

### DIFF
--- a/packages/wmr/src/wmr-middleware.js
+++ b/packages/wmr/src/wmr-middleware.js
@@ -39,7 +39,7 @@ export default function wmrMiddleware(options) {
 	distDir = resolve(dirname(out), distDir);
 
 	const NonRollup = createPluginContainer(getPlugins(options), {
-		cwd,
+		cwd: root,
 		writeFile: (filename, source) => writeCacheFile(out, filename, source),
 		output: {
 			// assetFileNames: '@asset/[name][extname]',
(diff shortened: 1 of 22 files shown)


JoviDeCroock (approved):

packages/wmr/src/wmr-middleware.js
```diff
@@ -39,7 +39,7 @@ export default function wmrMiddleware(options) {
 	distDir = resolve(dirname(out), distDir);
 
 	const NonRollup = createPluginContainer(getPlugins(options), {
-		cwd,
+		cwd: root,
```
developit, line 42: (meant to send this before) does this match what we initialize Rollup with in prod builds? It needs to match in order for paths to be the same I think (both will "work", but would generate different hashes).

marvinhagemeister: Just checked, there is no change in behaviour. Rollup itself doesn't seem to allow us to pass a custom `cwd` anywhere and here at this line we pass the same actual value as before. All plugins receive the same paths as before too.

PR #683, opened 2021-06-12
Use <link rel=modulepreload>

### LINKED ISSUE #682: Console warning: A preload for 'xyz.js' is found, but is not used because the script type does not match.
After creating a new project with `npm init wmr your-project-name` and starting the server with `npm start` or `npm run serve` I'm getting these console warnings (on Chrome and Firefox):

```
A preload for 'http://localhost:8080/index.24f43138.js' is found, but is not used because the script type does not match.

The resource http://localhost:8080/index.24f43138.js was preloaded using link preload but not used within a few seconds from the window's load event. Please make sure it has an appropriate `as` value and it is preloaded intentionally.
```

Is that OK?

developit:
Fixes #682.

### DIFF
--- a/.changeset/fresh-hounds-tease.md
+++ b/.changeset/fresh-hounds-tease.md
@@ -1,0 +1,5 @@
+---
+"create-wmr": patch
+---
+
+Use <link rel=modulepreload>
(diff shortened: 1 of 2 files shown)


JoviDeCroock (approved):

lukeed:
I can't remember exactly what it was, but does this mean the little performance hiccup of modulepreload is gone now?

marvinhagemeister (approved):

PR #684, opened 2021-06-13
Rewrite styles pipeline to get rid of transform abstraction for CSS

marvinhagemeister:
Completely rewrite our stylesheet pipeline. All CSS files can now be intercepted from plugins allowing us to leverage the same pipeline that we use for production during development.

- Fixes `.scss/.sass` not compiled on watch
- Fixes nested `.scss/.sass` files not compiled with sass
- Fixes `.scss/.sass` files not compiled when directly referenced in HTML
- Fixes asset caching overwriting each others entries
- Adds groundwork for intercepting urls inside css for aliasing (#685) or resolving from `node_modules` (didn't want to make this PR even larger, will come in a next PR)

Notes:

- Proxy modules `foo.module.css.js` are now specified by a query param: `foo.module.css?module`.
- Assets now have hashes to avoid collisions in our cache. We had assets that overwrote each others entries

This PR replaces #657 .

### DIFF
--- a/.changeset/shiny-adults-brake.md
+++ b/.changeset/shiny-adults-brake.md
@@ -1,0 +1,11 @@
+---
+'wmr': major
+---
+
+Completely rewrite our stylesheet pipeline. All CSS files can now be intercepted from plugins allowing us to leverage the same pipeline that we use for production during development.
+
+- Fixes `.scss/.sass` not compiled on watch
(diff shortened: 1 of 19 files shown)


JoviDeCroock (approved):

Inviz:
Yes, i'm so hot for this

PR #701, opened 2021-06-21
Fix crash when sass file not found

### LINKED ISSUE #700: Issues with sass resolver
**Describe the bug**
1) If import sass path is resolved to non-existing file, wmr can not recover from the error and crashes: `Error: ENOENT: no such file or directory, open '/Users/imac/Sites/medes/web/frontend/public/style/partials/variables.scss'` at `const contents = await fs.readFile(file, 'utf-8');`. I think it 
2) Relative @imports within aliased file paths do not point to rewritten location
3) "Bailing out if nothing changes" breaks the case with nested aliased file. I.e. there's alias on `style`, and then `style/index.scss` includes `style/another.scss`. Sass throws `Can't find stylesheet to import.`. Resolver thinks that "nothing changed" but dart-sass does not come up with the same resolved path somehow. Removing that condition fixes the issue

**To Reproduce**

1. Create a sass file that @imports file that gets rewritten but does not exist. Actually you won't need aliasing even, since resolver truncates absolute path to relative path and it is considered a rewrite by resolver.
2. Run wmr and load the page

**Expected behavior**
Wmr should not crash

marvinhagemeister:
This PR adds proper error handling around the sass compiler. In doing so one issue with potential incorrect file paths was found and resolved.

Fixes #700 .


packages/wmr/src/plugins/sass-plugin.js
```diff
@@ -62,7 +63,14 @@ export default function sassPlugin({ production, sourcemap, root }) {
 		// need a way to do the opposite here to support loading virtual
 		// sass modules. This is a limitation in Rollup. So for now we only
 		// do resolution here.
-		const resolved = await pluginResolve(url);
+		let resolved;
+		try {
+			resolved = await pluginResolve(url);
+		} catch (err) {
+			done(err);
```
marvinhagemeister, line 70: TODO: Add a test for this

packages/wmr/src/plugins/sass-plugin.js
```diff
@@ -74,8 +82,8 @@ export default function sassPlugin({ production, sourcemap, root }) {
 		if (!path.isAbsolute(file)) {
 			file = path.join(root, file);
 		}
-		const contents = await fs.readFile(file, 'utf-8');
-		done({ file, contents });
+
+		done({ file });
```
marvinhagemeister, line 86: This doesn't solve a bug, but there is no need for us to read the file ourselves. Sass supports either `contents` or `file`, but not both. If both are set it will fallback to `contents`.

marvinhagemeister:
Currently blocked on: https://github.com/sass/dart-sass/issues/1370

marvinhagemeister:
I'm starting to think that there isn't a good option to support aliases until sass's `importer` API is more stable. It's a hit and miss in both `sass` and `node-sass` right now. Switching to the `file` option like in this PR would break aliasing, but would make `relative` paths in sass work which seems to be the better default behaviour.

marvinhagemeister:
Alright, I think it's ready to be merged. It improves the sass integration substaintially by catching errors and fixing a nasty file path issue. Both scenarios are covered with a test in this PR 👍

Inviz:
Thanks! Will try it out as soon as it's released

marvinhagemeister:
@Inviz Just published WMR `3.1.0` 🎉

Inviz:
I think I'm still seeing the issue of `Relative @imports within aliased file paths do not point to rewritten location` i mentioned. An aliased file having relative imports, tries to import them off root

Inviz:
scss files linked via <link> in html don't get hot-reloaded, but they do if they are linked in .js file

marvinhagemeister:
@Inviz I think I was able to resolve the first issue, but I have trouble reproducting the other ones. On all the machines I have access to (Windows, macOS, Arch Linux), it applies HMR updates even if the `.scss` file is directly referenced in a link tag. I don't see the performance issue either, but maybe my test cases are too small. Every update, no matter the type, is instant.

Inviz:
Yeah, maybe i will need to spin up the new project for this link-tag thing and see what's up. Thanks for trying.

As for the last one it's not so much about performance, but more about literally only reloading file after it is saved 4 times. Quickly or slowly. Why 4? no idea.  I will look more into it and report if I will find any more details on this. Thanks!

marvinhagemeister:
Just checked with WMR `3.4.0` and the variables case you mentioned works with HMR 🎉

PR #715, opened 2021-06-28
Add groundwork for source map support

### LINKED ISSUE #699: Improve source map handling
The source map handling in WMR can be a lot better. Namely:

- [x] Ensure `--sourcemap` generates sourcemaps #715 
- [x] Merge source maps in `container.transform()`
- [x] Ensure `acorn-traverse` generates correct mappings
- [x] Always include `sourcesContent`, so that the original code is shown in browser devtools
- [ ] ~~Ensure npm plugin supports source maps (might need a rewrite)~~ - out of scope for now. The rewrite will happen either way and we can do the source maps in one go
- [x] Add source map support to `.sass` (iirc they have different file mappings)
- [x] Decide on path format `./foo.js` vs `foo.js`
- [x] Improve performance of source map generation (check if it can be made so fast, that we can turn this always on)

marvinhagemeister:
This PR adds a bit of groundwork for supporting source maps in WMR. There is lots more work to do, but this is a complete enough start ready to be merged and to build upon.

- [x] Fix plugin container not forwarding source maps
- [x] Fix invalid `null` value in `sources` array of source maps
- [x] Fix `.map` files not being served
- [x] Fix `options.sourcemap` not passed to plugins
- [x] Fix `options.sourcemap` always being `false`
- [x] Skip transform in `import-assertion` plugin when no changes were made.
- [x] Merge source maps in transform step
- [x] Update plugins to return source maps when they transform the code

Fixes #699

### DIFF
--- a/.changeset/mighty-ligers-beam.md
+++ b/.changeset/mighty-ligers-beam.md
@@ -1,0 +1,5 @@
+---
+'wmr': patch
+---
+
+Fix WMR not serving `.map` files and lay groundwork for source map support in `development` mode.
--- a/.changeset/mighty-ligers-beam.md
+++ b/.changeset/mighty-ligers-beam.md
@@ -2,4 +2,4 @@
 'wmr': patch
 ---
 
-Fix WMR not serving `.map` files and lay groundwork for source map support in `development` mode.
+Rewrite internal source map handling. This adds full support for source maps during `development` and `production` and ensures that `.map` files are served correctly.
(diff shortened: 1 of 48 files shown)


marvinhagemeister:
Currently waiting on https://github.com/Rich-Harris/sourcemap-codec/pull/85 to be merged and released. It's the reason our CI is failing here.

marvinhagemeister:
Alright, used `patch-package` to patch `sourcemap-codec` and `es-module-lexer` locally to support native ESM in node 12.18 and make our CI green again. Filed upstream PRs for both packages. Once they are merged and released we can easily drop our floating patches without hassle.

Waleed19915 (commented):
Finally

PR #720, opened 2021-06-29
Rename prerenderLinks to customRoutes

marvinhagemeister:
Follow up to #717


rschristian (approved):
Thanks for also fixing up the stuff that I forgot / didn't notice (jsdoc, actual docs, etc.)

PR #724, opened 2021-07-01
Add support for passing any file to plugins

### LINKED ISSUE #721: WMR does not support .mdx or .md files
## **Describe the bug**
`.mdx` as well as `.md` files are not being passed to a plugin specifically added in order to handle those file types. As a result, the user gets errors in their browser upon serving those files that have not been processed. 

## **To Reproduce**
```
yarn create wmr xdm-example
yarn add xdm --dev
```

Add an import to `public/index.js`
```js
import Content from './content.mdx';
```

Create `public/content.mdx`
```bash
echo "# Hello World" > public/content.mdx
```

Add XDM Rollup plugin
```js
import xdm from 'xdm/rollup.js';`

    config.plugins.push(xdm());
}
```

Navigate to `node_modules/xdm/lib/integrations/rollup.js` and add at line 32:
```js
console.log('xdm:', path);
```

Then start the dev server. Notice in the logs:
```
xdm: ./style.css
xdm: ./index.js
xdm: ./pages/home/index.js
xdm: ./pages/_404.js
xdm: ./header.js
```

No mention of `content.mdx`, and the browser should print the following error into the console:

```
Loading module from “http://localhost:3000/example.mdx” was blocked because of a disallowed MIME type (“text/mdx”).
```

## **Expected behavior**
MDX files (along with other file types as well) should be passed to plugins that might be used in order to handle those file types.

- Node Version: 14.16.0
- WMR Version: 3.2.0

## **Additional context**
I'll certainly look into it, but there's areas of WMR that I still haven't even glanced at, so not sure how far I'll get.

marvinhagemeister:
This PR adds support for passing any file forward to plugins. This allows plugins to intercept all kinds of files. The only exception to that are html files for now.

Fixes #721 .

### DIFF
--- a/packages/wmr/src/wmr-middleware.js
+++ b/packages/wmr/src/wmr-middleware.js
@@ -1,4 +1,4 @@
-import { resolve, dirname, relative, sep, posix, isAbsolute, normalize, basename } from 'path';
+import { resolve, dirname, relative, sep, posix, isAbsolute, normalize, basename, extname } from 'path';
 import { promises as fs, createReadStream } from 'fs';
 import * as kl from 'kolorist';
 import { getWmrClient } from './plugins/wmr/plugin.js';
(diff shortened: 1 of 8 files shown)


packages/wmr/src/wmr-middleware.js
```diff
@@ -284,7 +284,14 @@ export default function wmrMiddleware(options) {
...
 			cacheKey += '?asset';
 			transform = TRANSFORMS.asset;
-		} else if (prefix || hasIdPrefix || isModule || /\.([mc]js|[tj]sx?)$/.test(file) || /\.(css|s[ac]ss)$/.test(file)) {
+		} else if (
+			prefix ||
+			hasIdPrefix ||
+			isModule ||
+			/\.([mc]js|[tj]sx?)$/.test(file) ||
+			/\.(css|s[ac]ss)$/.test(file) ||
+			(extname(file) !== '' && extname(file) !== 'html' && !/favicon\.ico$/.test(file))
```
developit, line 293: I'm thinking maybe it would make sense to only intercept imports here? `import './foo.md'` is transformed to `import './foo.md?module'` and can be intercepted by plugins, but `fetch('/foo.md')` remains a simple static file.

```suggestion
			(isModule && extname(file) !== '' && extname(file) !== 'html' && !/favicon\.ico$/.test(file))
```

marvinhagemeister: The `isModule` check is already present a few lines above, but your snippet gave me an idea. With the recent commits we now append `?module` after transforming the `.js` files, similar to how we handle style imports.

PR #732, opened 2021-07-06
Extract Prefresh code into a separate plugin

marvinhagemeister:


developit (approved):
LGTM!

packages/wmr/src/plugins/preact/prefresh.js
```diff
@@ -0,0 +1,55 @@
+import MagicString from 'magic-string';
+import path from 'path';
+
+const PREFRESH = `
+if (import.meta.hot) {
```
developit, line 5:
```suggestion
import '@prefresh/core';
if (import.meta.hot) {
```

packages/wmr/src/plugins/preact/prefresh.js
```diff
@@ -0,0 +1,55 @@
...
+			// Only inject into modules with JSX and exports
+			if (!hasJsx || !hasExport) return;
+
+			const s = new MagicString(code, {
+				filename: id,
+				// @ts-ignore
+				indentExclusionRanges: undefined
+			});
+
+			s.prepend(`import '@prefresh/core';\n`);
```
developit, line 46:
```suggestion
```

developit:
The CI failure is interesting... it's not getting HMR injected for some reason, but I can't see how this PR would cause that to happen.

marvinhagemeister:
@developit The CI failure is caused by a commit of mine that was intended for another branch 😬 Dropping that makes it pass again.

PR #734, opened 2021-07-07
Fix invalid TypeScript ReturnType parameter

marvinhagemeister:
The type `ReturnType` only works on types, not on values. A function definition is by default a value and not a type, therefore we need to convert it to a type first via `typeof`.

This fixes intellisense for those type and the popup that shows up when you hover it.

// cc @developit


developit (approved):
Ah! At some point I started forgetting the typeof... I can only imagine I've done this elsewhere lol

PR #738, opened 2021-07-08
Fix flaky HMR test by waiting until full page load

marvinhagemeister:
This fixes flaky test runs on my windows machine


developit:
Ha! This must have taken forever to track down. Makes sense now that I see it though - the default `networkidle2` would potentially fire before the websocket connection has been set up, since it's often the last thing to happen.

PR #739, opened 2021-07-08
Fix invalid code generated by template helper

marvinhagemeister:
This PR makes both `template` and `template.ast` work without generating invalid code.

The problem occurs due to the optimization where we try to re-use the original code and slice output from that. The heuristic is based on whether a node has position data or not. But with `template`, `template.ast` or `t.clone()` it always has position data leading to wrong results.

The fix here adds a `forceRegenerate` flag that skips this optimization for `path.replaceWith`. The other option would be to visit each node of the replaced tree and delete the position data.

- Fixes crash when template string contained unrelated uppercase latters
- Fix invalid code generated by `path.replaceWith` when invoked with nodes that contain position data.
- Overwrite `ObjectExpression` codegen to be roughly more in line with expected formatting inside JSX or arrow function returns

### DIFF
--- a/packages/wmr/src/lib/acorn-traverse.js
+++ b/packages/wmr/src/lib/acorn-traverse.js
@@ -66,6 +66,25 @@ export function generate(node, ctx) {
 
 let codeGenerator = {
 	...astring.baseGenerator,
+	ObjectExpression(node, state) {
+		if (!node.properties.length) {
+			state.write('{}');
+		} else if (node.properties.length <= 5) {
+			state.write('{ ');
+			for (let i = 0; i < node.properties.length; i++) {
+				const prop = node.properties[i];
+				this[prop.type](prop, state);
+				if (i < node.properties.length - 1) {
+					state.write(', ');
+				}
+			}
+			state.write(' }');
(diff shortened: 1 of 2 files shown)


packages/wmr/src/lib/acorn-traverse.js
```diff
@@ -357,10 +388,21 @@ class Path {
 		if (this._regenerateParent()) {
 			this._hasString = false;
 		} else {
+			// Skip string generate optimizations as node positions won't
+			// match anymore.
+			this.ctx.forceRegenerate = true;
```
developit, line 393: Will this skip string optimizations on every replaceWith() call? Or only when encountering a template?

marvinhagemeister: This does skip every string optimization. The only other option I'm aware of would be to traverse the return value of `template/template.ast/t.clone()` and delete position data on every node.

marvinhagemeister: Alright went with clearing the node positions instead to keep the string optimization.

PR #744, opened 2021-07-08
[docs] Slide in sidebar when first shown

developit:
The sidebar comes in before the content when clicking "Get Started". This animation makes that feel a little nicer.

I was slightly worried about what happens on initial page load / reloads directly on `/docs` pages though.

### DIFF
--- a/docs/public/styles/index.css
+++ b/docs/public/styles/index.css
@@ -437,6 +437,13 @@ h5:hover .anchor {
 	z-index: 200;
 	outline: none;
 	transition: background 0.3s;
+	animation: sidebar-slide 300ms ease forwards 1 200ms;
+}
+
+@keyframes sidebar-slide {
+	0% {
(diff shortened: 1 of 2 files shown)


docs/public/styles/index.css
```diff
@@ -437,6 +437,13 @@ h5:hover .anchor {
 	z-index: 200;
 	outline: none;
 	transition: background 0.3s;
+	animation: sidebar-slide 500ms ease forwards 1;
```
marvinhagemeister, line 440: I like the animation a lot, but it's showing up on the home page too. There we don't have a sidebar. Maybe we just need to trigger the animation when we detect a route change away from home?

Screenshot when I've just got to the home page. Haven't navigated anywhere yet.

Screenshot 2021-07-08 at 18 05 48

marvinhagemeister (approved):

PR #762, opened 2021-07-11
Fix unable to intercept wmr middleware

### LINKED ISSUE #297: Proxy middleware can't be first in middleware-stack
I've added `http-proxy-middleware` to the middleware stack in `wmr.config.(m)js`, as I need to proxy certain routes to a backend. Unfortunately I can't make it work, and I think the problem is, that the proxy middleware needs to be first in the middleware stack.

My suggestion is to do one of the following
1) Split config.middleware into two arrays, one to apply before any other middlewares are applied, and one to apply after.
2) Add a proxy element to the config, where one can simply add proxy configurations, that will be applied to the polka server using e.g. `http-proxy-middleware` before any other middlewares are applied. I.e. something like

```
// wmr.config.(m)js
config.proxy = {
  "/api": {
    changeOrigin: true,
    ws: true,
    logLevel: "debug",
    target: "http://my.fancy.backend",
    pathRewrite: {
      "^/api": "",
    },
  }
};
```

I've made a small poc of the latter and would happily make a PR on it.

marvinhagemeister:
Similar to custom plugins, custom middlewares are now handled before our own internal ones. This allows requests to be intercepted which wasn't possible before.

Fixes #297 .

### DIFF
--- a/packages/wmr/src/server.js
+++ b/packages/wmr/src/server.js
@@ -104,6 +104,11 @@ export default async function server({ cwd, root, overlayDir, middleware, http2,
 		app.use(compression({ threshold, level: 4 }));
 	}
 
+	// Custom middlewares should always come first, similar to plugins
+	if (middleware) {
+		app.use(...middleware);
+	}
+
 	app.use('/@npm', npmMiddleware({ alias, optimize, cwd }));
 
 	// Chrome devtools often adds `?%20[sm]` to the url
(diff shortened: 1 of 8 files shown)


packages/wmr/src/server.js
```diff
@@ -98,6 +98,11 @@ export default async function server({ cwd, root, overlayDir, middleware, http2,
 
 	app.ws = new WebSocketServer(app.server, '/_hmr');
 
+	// Custom middlewares should always come first, similar to plugins
```
developit, line 101: I'm wondering if we should have custom middleware come after the compression middleware. It modifies `res.write` & friends, but doing so after they've been called might break things.

marvinhagemeister: Good point, updated the PR

PR #764, opened 2021-07-11
Add support for scope tracking

marvinhagemeister:
This lays the foundation for scope tracking in our acorn transformation layer. It includes a better fix for the `parentPath` issue than #763 which is coincidentally tested with the scope tracking stuff.

_FYI: This doesn't implement the second parsing pass that's needed to collect all references in scope, that will come best with the next PR_

### DIFF
--- a/packages/wmr/src/lib/acorn-traverse.js
+++ b/packages/wmr/src/lib/acorn-traverse.js
@@ -347,14 +347,6 @@ class Path {
 		let ancestors = this.ancestors.slice();
 		let parent = ancestors.pop();
 		if (!parent) return undefined;
-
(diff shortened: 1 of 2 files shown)


packages/wmr/src/lib/acorn-traverse.js
```diff
@@ -401,7 +393,8 @@ class Path {
 			const name = token[1] || token[2];
 			const prev = node;
 			node = node[name];
-			if (!Array.isArray(node)) {
+
+			if (!Array.isArray(prev)) {
```
marvinhagemeister, line 397: This here is the proper fix for the `parentPath/parent` issue.

PR #774, opened 2021-07-20
Fix absolute root-relative imports not working

### LINKED ISSUE #773: Absolute includes dont work in (s)css + root inclusion doesn't work in build process
**Describe the bug**
Scss files that include other files as absolute path fail to compile (Cant find stylesheet to import).
Additionally in build absolute paths don't work when javascript files include each other.

**To Reproduce**
https://github.com/Inviz/wmr-issues/tree/main/3.root-in-scss

CSS:
1. Create two scss files in `public`
2. Make 1st file `@include /2.scss`
3. Include 1st file in js

JS: 
1. Use `import "/absolute.js"` in js
2. Works in dev
3. Breaks in build (could not be resolved)

**Expected behavior**
Root inclusion should work in scss just as in javascript, build and dev

**Bug occurs with:**

- [ ] `wmr serve`

marvinhagemeister:
Fixes #773

### DIFF
--- a/.changeset/yellow-cars-tie.md
+++ b/.changeset/yellow-cars-tie.md
@@ -1,0 +1,5 @@
+---
+'wmr': patch
+---
+
+Fix absolute imports starting with `/` not working in SASS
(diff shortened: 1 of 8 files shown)


Inviz:
This also solves absolute javascript includes in build, right? Didn't understand that from the PR.

Issue #771, opened 2021-07-20
SCSS is served uncompiled on build if linked from html file


Inviz:
**Describe the bug**
`<link rel="stylesheet" href="style.scss" />` works in dev, but not on production. It is served as is.

**To Reproduce**

1. Rename style.css to style.scss
2. Re-link it in index.html
3. Use something of sass like varible
4. Try in dev: works: scss compiled
5. Try in build: doesn't: served as is

Reproduction project: https://github.com/Inviz/wmr-issues/tree/main/1.sass-in-html

**Expected behavior**
Scss should never be served uncompiled

**Bug occurs with:**

- [ ] `wmr` or `wmr start` (development)
- [x] `wmr build` (production)
- [x] `wmr serve`

PR #782, opened 2021-07-27
Transpile scss files included from html

JoviDeCroock:
Fixes: https://github.com/preactjs/wmr/issues/771

We were emitting the asset but rollup does not push that through the plugin-pipeline, hence why this is now happening as a post-step when generating the final bundle.


Inviz:
Looking forward to this

PR #783, opened 2021-07-27
Handle JSON string source maps returned from rollup plugins

Aeriit:
Some rollup plugins (such as [rollup-plugin-esbuild](https://www.npmjs.com/package/rollup-plugin-esbuild)) return the source map as a string instead of an object; this string needs to be parsed before we can normalize the source URLs (because right now it throws a "cannot read property 'map' of undefined").

### DIFF
--- a/packages/wmr/src/lib/rollup-plugin-container.js
+++ b/packages/wmr/src/lib/rollup-plugin-container.js
@@ -290,6 +290,18 @@ export function createPluginContainer(plugins, opts = {}) {
 
 				logTransform(`${kl.dim(formatPath(id))} [${plugin.name}]`);
 				if (typeof result === 'object') {
+					if (typeof result.map === 'string') {
+						try {
+							result.map = JSON.parse(result.map);
+						} catch {
+							if (hasDebugFlag()) {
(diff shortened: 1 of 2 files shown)


packages/wmr/src/lib/rollup-plugin-container.js
```diff
@@ -290,6 +290,18 @@ export function createPluginContainer(plugins, opts = {}) {
 
 				logTransform(`${kl.dim(formatPath(id))} [${plugin.name}]`);
 				if (typeof result === 'object') {
+					if (typeof result.map === 'string') {
+						try {
+							result.map = JSON.parse(result.map);
+						} catch {
+							if (hasDebugFlag()) {
```
marvinhagemeister, line 297: No need to wrap this with `hasDebugFlag()`. That check is already part of `logTransform`.

marvinhagemeister (commented):
Looking good! Let's add a test case for this so we don't run the risk of regressing on this behavior here.

For that you can create a new sample project in the `fixture` directory and add a simple noop plugin, that returns a `string` for the `map` property. There are a few examples on how to run those fixtures from our test suite in  ´fixtures.test.js`

PR #785, opened 2021-07-28
[feature] Automatically reconnect to HMR with exponential backoff

developit:


packages/wmr/src/plugins/wmr/client.js
```diff
@@ -31,6 +33,15 @@ function connect(needsReload) {
 
 	ws.addEventListener('message', handleMessage);
 	ws.addEventListener('error', handleError);
+	ws.addEventListener('close', reconnect);
```
JoviDeCroock, line 36: We could also listen for something like focussing the browser again

Inviz: Yes, that's a great idea. Thtat's what I'm using to run my code for waking up ios wrapper app

Inviz: since the exponential dropoff with 30s limit will not solve the case for a sleep longer than 30s, but visibilitychange certainly would.

Thanks guys!

packages/wmr/src/plugins/wmr/client.js
```diff
@@ -31,6 +41,23 @@ function connect(needsReload) {
...
+}
+
+function reconnect() {
```
JoviDeCroock, line 47: Currently reconnect only gets called once, should this be called again if establishing connection in `connect(false)` fails?

PR #786, opened 2021-07-31
(wmr/hmr) - implement focus-based reconnect

JoviDeCroock:


developit:
@JoviDeCroock I'm wondering if we need to combine the timer and focus checks? Reconnecting on focus/visibilitychange won't address the case where network connection drops/changes while something is in the foreground. 

It seems like we could start reconnecting if the documents visibility state is visible, and also if the state changes to visible, in both cases triggering the exponential backoff approach?

marvinhagemeister (approved):
Let's add a changeset + the timer checks on top and then we're good to go 👍

PR #787, opened 2021-08-03
(wmr) - use swc to minify code

JoviDeCroock:


kdy1:
Can you try last version of swc?
Rules are not implemented yet, but now it generates much compact code.

JoviDeCroock:
Bumped it to latest

kdy1:
Seems like I made a mistake. I'll investigate

marvinhagemeister:
FYI: Managed to trim down the CI inifinite loop error to a minimal repro. See: https://github.com/swc-project/swc/issues/2044

marvinhagemeister:
The CI failure is caused by an upstream bug in `swc`'s minifier in regards to mangling variables and shorthand properties. Reported it here: https://github.com/swc-project/swc/issues/2051

marvinhagemeister:
FYI: Tests are failing due to a minification bug in swc: https://github.com/swc-project/swc/issues/2078

marvinhagemeister:
@JoviDeCroock looks like we'd only need to update the assertion to check for double quotes in the two failing tests 🎉

kdy1:
@JoviDeCroock Is there a reason you closed this?

JoviDeCroock:
@kdy1 well if I understood correctly the emphasis of swc has moved to being Rust-based for plugins and we don't currently have any Rust in wmr (and I don't know how that boats with my fellow contributors). I've closed this as I'm investigating what is causing the build to fail and want to isolate the bugs.

kdy1:
Minifier is not affected by plugin api. Plugin api only applies to `transform` apis.
Maybe https://github.com/swc-project/swc/issues/2262 can be the cause.

JoviDeCroock:
Currently seeing:

```
12:01:52 PM: The following error was thrown during prerendering:
12:01:52 PM: SyntaxError: Missing initializer in const declaration
12:01:52 PM: 
12:01:52 PM:     at Loader.moduleStrategy (internal/modules/esm/translators.js:81:18)
12:01:52 PM:     at async link (internal/modules/esm/module_job.js:37:21)
12:01:52 PM: /bin/sh: 1: wmr: not found
12:01:52 PM: error Command failed with exit code 127.
```

For the docs site, I'll look into it later

kdy1:
Seems like the minifier tries to inline or drop the initializer of the variable, but didn't remove the variable itself.
I'll try fixing some known issues of the minifier.

fabiosantoscode:
Love this ❤️

PR #788, opened 2021-08-06
(wmr) - add test for scss in html

JoviDeCroock:


packages/wmr/test/production.test.js
```diff
@@ -309,6 +309,13 @@ describe('production', () => {
 			});
 		});
 
+		it('should pick up scss from the html file', async () => {
+			await loadFixture('css-sass-html', env);
+			instance = await runWmr(env.tmp.path, 'build');
+			const dir = await fs.readdir(path.join(env.tmp.path, 'dist', 'assets'));
+			expect(dir.some(x => x.endsWith('.css'))).toBeTruthy();
```
marvinhagemeister, line 316: Looking good! Let's add a check on top of this, that verifies that the page loads and that the CSS is applied correctly.

```suggestion
			expect(dir.some(x => x.endsWith('.css'))).toBeTruthy();

			const code = await instance.done;
			expect(code).toEqual(0);
			const { address, stop } = serveStatic(path.join(env.tmp.path, 'dist'));
			cleanup.push(stop);
			await env.page.goto(address, {
				waitUntil: ['networkidle0', 'load']
			});
			expect(await env.page.$eval('div', el => getComputedStyle(el).color)).toBe('rgb(255, 0, 0)');
```

Issue #772, opened 2021-07-20
CSS cache-busting  suffix only takes first file into account


Inviz:
**Describe the bug**
Build process glues all css used in javascript together, but assigns name based on the first file used. If that file hasn't changed, the chunk name never updates leading to stale cache for people who revisit the app after update.

**To Reproduce**
Repo: https://github.com/Inviz/wmr-issues/tree/main/2.css-asset-stale-suffix

1. Link two css/sass files inside javascript like index.js
2. Run build process, note the filename is `first_file.deadbeef.css`
3. Change second file
4. Run build process, filename is still `first_file.deadbeef.css`

**Expected behavior**
If any of concatenated css files are changed, chunk name should reflect that and be regenerated

**Bug occurs with:**

- [ ] `wmr` or `wmr start` (development)
- [x] `wmr build` (production)
- [ ] `wmr serve`

Inviz:
I tried to work around this by rewriting the path and adding extra ?parameters to bust cache, but that makes runtime to re-insert the file with original path.

Inviz:
<img width="441" alt="Screen Shot 2021-07-23 at 20 47 53" src="https://user-images.githubusercontent.com/2708/126783773-e091a1a5-423e-4073-883b-60a69319e55c.png">

Issue #361, opened 2021-02-22
Unprocessed scss file gets copied over to dist during build


zgoda:
I have local font definitions imported from `style/fonts.scss` in `style/main.scss`.

```console
$ tree public
public
├── app.js
├── assets
│   └── fonts
│       ├── lato-v17-latin-ext_latin-italic.woff
│       ├── lato-v17-latin-ext_latin-italic.woff2
│       ├── lato-v17-latin-ext_latin-regular.woff
│       └── lato-v17-latin-ext_latin-regular.woff2
├── index.html
├── index.js
└── style
    ├── fonts.scss
    └── main.scss
```

Build properly builds css and puts content into single minified file, but leaves unprocessed `style/fonts.scss` in `dist`:

```console
$ tree dist
dist
├── assets
│   ├── fonts
│   │   ├── lato-v17-latin-ext_latin-italic.woff
│   │   ├── lato-v17-latin-ext_latin-italic.woff2
│   │   ├── lato-v17-latin-ext_latin-regular.woff
│   │   └── lato-v17-latin-ext_latin-regular.woff2
│   └── main.e75a073a.css
├── chunks
│   └── prerender.8ab96da0.js
├── index.13d2992f.js
├── index.html
└── style
    └── fonts.scss
```

JoviDeCroock:
Hey,

How do these files relate, does main.scss import the fonts file or? Need to know this if we want to reproduce this ourselves.

rschristian:
Duplicate of #254 I believe

zgoda:
main.scss:

```
@import './fonts.scss';
```

zgoda:
In case of SCSS the import is properly resolved by SASS upon processing and the resulting css file is perfectly valid.

rschristian:
@zgoda So just a leftover file then? Functionally everything is working as you'd expect?

zgoda:
Apart of https://github.com/preactjs/wmr/issues/291 evetything seems to work fine. ;)

PR #789, opened 2021-08-07
several sass fixes

JoviDeCroock:
- [x] Does not duplicate assets
- [x] When running build twice the hash doesn't update
- [x] Using an scss file included from html does not support imports

Fixes https://github.com/preactjs/wmr/issues/772
Fixes https://github.com/preactjs/wmr/issues/361

### DIFF
--- a/packages/wmr/src/lib/plugins.js
+++ b/packages/wmr/src/lib/plugins.js
@@ -38,6 +38,7 @@ export function getPlugins(options) {
 	if (split === -1) split = plugins.length;
 
 	const production = mode === 'build';
+	const mergedAssets = new Set();
 
 	return [
 		acornDefaultPlugins(),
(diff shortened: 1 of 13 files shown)


packages/wmr/src/plugins/html-entries-plugin.js
```diff
@@ -97,7 +98,16 @@ export default function htmlEntriesPlugin({ root, publicPath }) {
 					all.push(ref);
 					waiting.push(
 						fs.readFile(abs, 'utf-8').then(source => {
-							this.setAssetSource(ref, source);
+							if (/\.s[ac]ss$/.test(abs)) {
+								transformSass.call(this, abs, abs, source, root, true, sourcemap).then(result => {
```
JoviDeCroock, line 102: Moved the `generateBundle` call from `sass-plugin` here as this ensures that we resolve imported files correctly during transformation. Else we emit the scss asset here as an asset and will be transforming in the dist-folder

packages/wmr/src/lib/plugins.js
```diff
@@ -38,12 +38,13 @@ export function getPlugins(options) {
 	if (split === -1) split = plugins.length;
 
 	const production = mode === 'build';
+	const mergedAssets = new Set();
```
JoviDeCroock, line 41: This file tracks assets that aren't relevant to the final build, for instance sass-plugin inlining them

marvinhagemeister (approved):
Fanatastic work 🙌 🏅

Issue #800, opened 2021-08-25
"File not found: undefined" on a brand new project


peterbe:
**Describe the bug**
Can't go to <http://localhost:8080> after having `npm start`. 

<img width="1143" alt="Screen Shot 2021-08-25 at 11 04 39 AM" src="https://user-images.githubusercontent.com/26739/130816090-a21f4dde-82a4-45d1-ac2c-a812a8bf3638.png">

**To Reproduce**

1. `cd /tmp`
2. `npm init wmr repro-overlay-bug`
3. ` cd repro-overlay-bug`
4. `npm start`
5. `open http://localhost:8080`

Note that I haven't touched any files at all. Just run `npm init` and `npm start`. 

**Expected behavior**
I can actually see the sample Home page that is built for a split second but then it goes away. 
Running `npm run build` and `PORT=8080 npm run serve` followed by going to <http://localhost:8080> works fine. 

**Bug occurs with:**

- [x] `wmr` or `wmr start` (development)
- [ ] `wmr build` (production)
- [ ] `wmr serve`

**Desktop (please complete the following information):**

- OS: macOS 11.5
- Browser: latest chrome and firefox
- Node Version: v16.3.0
- WMR Version: 3.5.1

**Additional context**

We really should have a GitHub Action that tests `npm start` and going to the sample app in debug mode. It would have caught this bug.

marvinhagemeister:
Just checked on my linux box and there it works fine. Maybe it's an OS dependent thing?

rschristian:
Or an existing attached service worker perhaps?

marvinhagemeister:
Fixed with #802 . We'll cut a new release for wmr soon.

PR #801, opened 2021-08-25
playwright test the demo

peterbe:
When I filed https://github.com/preactjs/wmr/issues/800 I thought I had discovered a bug in the demo app, which is a great app by the way. 
Turns out it was a stale service worker from another (preactcli) app I was testing also on `localhost:8080`. Anyway, I thought we could benefit from a full end-to-end test that checks that the demo always works. 
I have no idea if this is useful to the project or even desired at all. I like the idea of using `playwright` to spin up the demo and viewing it with a real browser.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -56,6 +56,38 @@ jobs:
         if: ${{ needs.changes.outputs.preact-iso == 'true' }}
         run: yarn eslint packages/preact-iso && yarn iso test
 
+  demo:
+    needs: changes
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        # Change the condition for ESM Dist Test below when changing this.
+        node-version: [12.18, 14.x]
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Use Node.js ${{ matrix.node-version }}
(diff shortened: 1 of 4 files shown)


developit:
I've never used Playwright - is it better suited to automated UI testing than what we're using now?

marvinhagemeister:
> I've never used Playwright - is it better suited to automated UI testing than what we're using now?

Yup, `puppeteer` is more like "here is a headless browser and that's it", whereas `playwright` has the philosophy: "here is everything you might need for doing browser-based testing, inspection and debugging". They allow you to capture a [full trace of each test](https://playwright.dev/docs/trace-viewer) that can be used to inspect what happened _after_ the test has failed. This is useful if for debugging CI failures.

But that being said and if I'm honest, we don't really need UI testing in WMR as we don't have any UI components to test. Our main use case for a headless browser is checking if a stylesheet or script loaded, that's it. That's why I'm neutral to both tools. They both work and I doubt there is much of a difference for testing WMR specifically. If we would test any actual UI stuff than `playwright` would be the clear winner for sure.

marvinhagemeister:
Hope you're ok if we close this PR as we have e2e tests that cover the functionality that's tested in this PR.

PR #806, opened 2021-08-29
Add support for compiling `.less` stylesheets

marvinhagemeister:
This PR adds support for transpiling `.less` stylesheets in the same way we do with sass.


developit:
Nice - looks like this does the thing where we will use `less` if installed, otherwise warn?

marvinhagemeister:
@developit yup, same way as we do with sass.

PR #810, opened 2021-08-29
Fix unable to load `.json` files outside of public directory

marvinhagemeister:


packages/wmr/src/plugins/json-plugin.js
```diff
@@ -15,37 +13,25 @@ import { promises as fs } from 'fs';
...
-
-			id = id.slice(IMPORT_PREFIX.length);
+			if (id.startsWith(IMPORT_PREFIX)) {
+				id = id.slice(IMPORT_PREFIX.length);
+			}
+			if (!id.endsWith('.json')) return;
 
 			const resolved = await this.resolve(id, importer, { skipSelf: true });
-			return resolved && INTERNAL_PREFIX + resolved.id;
+			return resolved && resolved.id;
```
developit, line 26: A thought: what if we returned `${resolved.id}?json` here? (or `${resolved.id}?module`?)

marvinhagemeister: We currently add `?module` to any script file that's imported from a script. I feel like that's a more reliable approach as this allows us to differentiate urls that were fetched directly via `fetch()` calls vs urls that were loaded by an import statement.

marvinhagemeister:
Hmm, I'm running into some issues under windows with these changes. Will investigate.

PR #819, opened 2021-08-30
Fix process.env import skewing env inlining offsets

### LINKED ISSUE #818: SyntaxError when loading script with process.env
**Describe the bug**
It appears there is a bug in one of the development transforms that incorrectly transforms the following code:
```js
let FORCE_COLOR, NODE_DISABLE_COLORS, NO_COLOR, TERM;
if (typeof process !== "undefined") {
  ({ FORCE_COLOR, NODE_DISABLE_COLORS, NO_COLOR, TERM } = process.env);
}

```
to
```js
var process={"browser":true,"env":{"NODE_ENV":"development"}};
let FORCE_COLOR, NODE_DISABLE_COLORS, NO_COLOR, TERM;
if (typeof process !== "undefined") {
  ({ FORCE_COLOR, NODE_DISAtypeof undefined COLOR, TERM } = process.env);
}
```

Removing the `typeof process !== 'undefined'` check results in the bug going away. This code can be find in the wild in the kleur package: https://github.com/lukeed/kleur/blob/598f24cb7f5054b7c504e9ab89251305311131b1/colors.mjs#L1

**To Reproduce**

1. Create the following files in a directory:
   ```html
   <!-- index.html -->
   <html>
   <script type="module" src="./index.js"></script>
   </html>
   ```
   ```js
   // index.js
   let FORCE_COLOR, NODE_DISABLE_COLORS, NO_COLOR, TERM;
   if (typeof process !== "undefined") {
      ({ FORCE_COLOR, NODE_DISABLE_COLORS, NO_COLOR, TERM } = process.env);
   }
   ```
2. Run `npx wmr`
3. Load http://localhost:8080/index.html
4. See the following error in the console: `Uncaught SyntaxError: Unexpected identifier`

Also

2. Run `npx wmr build`
3. Notice error the console:
   ```js
   Error: Unexpected token

    at Cs (D:\sandbox\wmr-parse-error\node_modules\wmr\wmr.cjs:2:138696)
    at va.error (D:\sandbox\wmr-parse-error\node_modules\wmr\wmr.cjs:2:177314)
    at va.tryParse (D:\sandbox\wmr-parse-error\node_modules\wmr\wmr.cjs:2:185502)
    at va.setSource (D:\sandbox\wmr-parse-error\node_modules\wmr\wmr.cjs:2:182924)
    at Gu.addModuleSource (D:\sandbox\wmr-parse-error\node_modules\wmr\wmr.cjs:2:353239)
    at async Gu.fetchModule (D:\sandbox\wmr-parse-error\node_modules\wmr\wmr.cjs:2:354416)
    at async Promise.all (index 0)
   ```

**Expected behavior**
WMR should properly transform the module.

**Bug occurs with:**

- [ ] `wmr serve`

- OS: Windows 10
- Browser: Edge 94
- Node Version: 14.17.4
- WMR Version: 3.6.0

developit:
This fixes #818. The issue was happening because `code` is passed to MagicString, which assumes all offsets are relative to that string. However, `code` was being modified to prepend an import statement before it gets scanned by the regex, which caused all offsets for `.overwrite()` to be relative to the modified `code` instead of the one that was passed to MagicString.

### DIFF
--- a/.changeset/dry-forks-yell.md
+++ b/.changeset/dry-forks-yell.md
@@ -1,0 +1,5 @@
+---
+"wmr": patch
+---
+
+Fix complex `process.env` usage (ex: `let {A}=process.env`, but not `process.env.A`) generating invalid code.
--- a/packages/wmr/src/plugins/process-global-plugin.js
+++ b/packages/wmr/src/plugins/process-global-plugin.js
@@ -69,7 +69,6 @@ export default function processGlobalPlugin({ NODE_ENV = 'development', env = {}
 			});
 
 			code = result.code;
-
 			const s = new MagicString(code);
 
 			// if that wasn't the only way `process.env` was referenced...
@@ -77,10 +76,8 @@ export default function processGlobalPlugin({ NODE_ENV = 'development', env = {}
 				// hack: avoid injecting imports into commonjs modules
 				if (/^\s*(import|export)[\s{]/gm.test(code)) {
 					s.prepend(`import process from '${PREFIX}process.js';\n`);
-					code = `import process from '${PREFIX}process.js';${code}`;
 				} else {
 					s.prepend(`var process=${processObj};\n`);
-					code = `var process=${processObj};${code}`;
 				}
 			}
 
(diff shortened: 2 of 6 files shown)


andrewiggins (approved):
Verified it fixes the issue for me locally!

marvinhagemeister (approved):
Good catch with the offsets 🙌

PR #832, opened 2021-09-02
Add support for partial route matching via `/foo/*`

marvinhagemeister:


packages/preact-iso/router.js
```diff
@@ -37,6 +37,8 @@ export const exec = (url, route, matches) => {
 		val = url[i];
 		// segment match:
 		if (!m && param == val) continue;
+		// /foo/* match
+		if (val && flag == '*') continue;
```
developit, line 41: I think this needs to apply only if there's no match name? (eg: `/foo/bar*`)

```suggestion
		if (!m && val && flag == '*') continue;
```

marvinhagemeister: Added your suggestion, but for some reason GitHub's UI think it's not part of the changeset 🤷

PR #839, opened 2021-09-04
Fix nested worker not found during prod builds

marvinhagemeister:


packages/wmr/src/plugins/worker-plugin.js
```diff
@@ -104,7 +104,8 @@ export function workerPlugin(options) {
 
 						const ref = this.emitFile({
 							type: 'chunk',
-							id: spec
+							id: spec,
+							importer: id
```
developit, line 108: TIL

PR #843, opened 2021-09-05
Sort list of output files by size within a group

marvinhagemeister:
Before:

Screenshot 2021-09-05 at 09 53 39

After:

Screenshot 2021-09-05 at 09 53 26


developit:
Awesome. I have wanted this since we did the asset sizes.

PR #851, opened 2021-09-05
Improve CLI output for compilation warnings

marvinhagemeister:
Instead of just showing the warning message we'll now show where the warning came from and potentially display a code frame of the code if available.

Before:

Screenshot 2021-09-05 at 18 36 47

After:

Screenshot 2021-09-05 at 18 31 59


developit (approved):
This looks great.

PR #852, opened 2021-09-05
Add code frame to uncaught exceptions if possible

marvinhagemeister:
- 174e5f6 Fix weird code frame formatting if line is way longer than terminal width
- 1d9791f Add code frame to uncaught exceptions if possible
- 859be34 Mark prerender errors as such when logging to the terminal

Interestingly all uncaught exceptions from a worker land up in the handler on the main thread. This happens often on my end during prerendering when a variable is only available in the browser, but not in node.

Need to find a way to catch those in the worker or somehow annotate them as originating from prerendering. But for now the code frame alone makes it a lot easier to understand what's going on.

Before:

Screenshot 2021-09-06 at 00 57 01

After:

Screenshot 2021-09-06 at 02 49 13

### DIFF
--- a/.changeset/early-carrots-appear.md
+++ b/.changeset/early-carrots-appear.md
@@ -1,0 +1,5 @@
+---
+'wmr': patch
+---
+
+Add code frame to uncaught exceptions if possible
--- a/packages/wmr/package.json
+++ b/packages/wmr/package.json
@@ -83,7 +83,7 @@
 		"sade": "^1.7.3",
 		"sass": "^1.34.1",
 		"semver": "^7.3.2",
-		"simple-code-frame": "^1.1.1",
+		"simple-code-frame": "^1.2.0",
 		"sirv": "^1.0.6",
 		"sourcemap-codec": "^1.4.8",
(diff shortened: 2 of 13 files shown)


developit:
I believe the "accepted" way to do this would be to use `process.on('uncaughtException')` in the Worker thread, and grab all the properties of the error there so they're available in the `error` event (by default it's just the `.message` property).

marvinhagemeister:
@developit from what I can gather the `onerror` callback is not called when an uncaught exception occurs. This seems to be the case with unknown variables.

developit:
@marvinhagemeister yeah a syntaxerror like undeclared var would bypass this because it won't hit the `catch()` callback. I can't think of a way to catch syntax errors without using eval or writing a temporary file.

marvinhagemeister:
@developit the `try/catch` seems to work 🎉

developit:
Nice, that's awesome. I'm guessing it must be thrown synchronously by `new Worker()` or something.

developit (approved):

PR #862, opened 2021-09-08
Don't flood terminal with Rollup's this is undefined warning

marvinhagemeister:
It's harmless and mainly caused by polyfills of libraries in
`node_modules`. Rollup deals with that automatically. Came across this with `monaco-editor` where every file seemingly includes polyfills. This leads to the terminal being spammed constantly.


developit:
Hah! I always wanted to find a way to turn those off.

Issue #836, opened 2021-09-03
`onLoadEnd` fires on every render on `preact-iso`'s Router


marvinhagemeister:
**Describe the bug**
`onLoadEnd` is fired on every render instead of only once when the promise is resolved.

**To Reproduce**

1. Render from a child under the router and update parent state above the router

**Expected behavior**
Only call `onLoadEnd` once the promise is flushed.

**Bug occurs with:**

- [x] `wmr` or `wmr start` (development)
- [x] `wmr build` (production)
- [ ] `wmr serve`

developit:
ahhh interesting - there are cases where routes don't suspend at all and `onLoadEnd` needs to get called (even though onLoadStart isn't called), but a state update that doesn't modify the route seems like it shouldn't re-invoke routing at all. Good catch.

marvinhagemeister:
Maybe we can just track if `onLoadEnd` was called before and skip calling that if it hasn't been invalidated by another `onLoadStart`

PR #864, opened 2021-09-08
(preact-iso) - avoid calling onLoadEnd on every render

JoviDeCroock:
Fixes https://github.com/preactjs/wmr/issues/836

Ref solution: https://github.com/preactjs/wmr/pull/864/commits/d6c87e340538c341e7826b7b043dec03ddc306fd (ties into the effect on every render)
State solution: https://github.com/preactjs/wmr/pull/864/commits/960364758bd818d8c00a482ef4c346c7adc05554 (avoids using effect on every render)

### DIFF
--- a/packages/preact-iso/router.js
+++ b/packages/preact-iso/router.js
@@ -87,14 +87,14 @@ export function LocationProvider(props) {
 }
 
 const RESOLVED = Promise.resolve();
-
 export function Router(props) {
-	const [, update] = useReducer(c => c + 1, 0);
+	const [c, update] = useReducer(c => c + 1, 0);
 
 	const { url, query, wasPush, path } = useLocation();
 	const { rest = path, params = {} } = useContext(RouteContext);
 
 	const isLoading = useRef(false);
+	const prevRoute = useRef(path);
 	// Monotonic counter used to check if an un-suspending route is still the current route:
 	const count = useRef(0);
 	// The current route:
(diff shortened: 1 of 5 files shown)


packages/preact-iso/router.js
```diff
@@ -176,7 +178,8 @@ export function Router(props) {
 
 		// The route is loaded and rendered.
 		if (wasPush) scrollTo(0, 0);
-		if (props.onLoadEnd) props.onLoadEnd(url);
+		if (props.onLoadEnd && isLoading.current) props.onLoadEnd(url);
+		isLoading.current = false;
 	});
```
JoviDeCroock, line 183: The bigger issue is here as we're basically calling this `layout-effect` on _ever single render_ rather than when it's needed, as this effect is inherently built on refs it's hard to give a reliable deps array, a reliable way would be to use a `state` for the loading times

developit: Hmm - that behavior is intentional though - the effect callback runs after every render to check whether that render suspended or not.

developit:
If we land this, I feel like we should also land a callback that fires after every successful route (sync or async), like `onDidRoute`. I often want to register a callback to do stuff like change title/reset scroll/etc

packages/preact-iso/router.js
JoviDeCroock: Yes but suspending should be part of the observed state else you're hoping that the render cycle of your top-level router ties in with your suspended boundary, this implies that your `useReducer` and `lazy.setState` happen correctly

marvinhagemeister (approved):
This is awesome! Love the tests! Thank you so much 🙌

PR #870, opened 2021-09-17
[preact-iso] Fix Router un-rendering incoming same-component route

developit:
I'm not 100% sure if this is the right solution, but reverting this line fixes the issue:
https://github.com/preactjs/wmr/pull/835/files#diff-b8c8c162d167c1ad31cb8461c49197c9c5527c58c2cde64d5c082d7359878ffdL123

My thinking is that `url` here is the full `location.path + location.search`, so the worst case is that we're potentially re-rendering routers that are unaffected by a given URL change.

### DIFF
--- a/.changeset/rude-moons-grow.md
+++ b/.changeset/rude-moons-grow.md
@@ -1,0 +1,5 @@
+---
+"preact-iso": patch
+---
+
+- Fix `<Router>` accidentally un-rendering incoming routes that happen to resolve to the same component as the outgoing route.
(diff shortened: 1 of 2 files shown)


.changeset/rude-moons-grow.md
```diff
@@ -0,0 +1,5 @@
+---
+"preact-iso": patch
+---
+
+- Fix `<Router>` accidentally un-rendering incoming routes that happen to resolve to the same component as the outgoing route.
```
JoviDeCroock, line 5:
```suggestion
Fix `<Router>` accidentally un-rendering incoming routes that happen to resolve to the same component as the outgoing route.
```

JoviDeCroock (approved):

marvinhagemeister (approved):

PR #873, opened 2021-09-19
(chore) - speed up ci execution by seperating the steps

JoviDeCroock:
This splits up the tests for preact-iso, wmr-dev and wmr-prod so we have a faster output across more machines.

If we merge this we'll have to tweak the mandatory status checks

CC @marvinhagemeister should we add 16.x to our matrix?

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -31,7 +31,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [12.18, 14.x]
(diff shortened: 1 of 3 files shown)


marvinhagemeister (approved):
This is awesome!

+1 on adding Node 16 as that's what every M1 chip user uses. Looking at our code I don't expect any difference in test outcome compared to Node 14 though.

JoviDeCroock:
@marvinhagemeister it errors out on Node 16 so I'll leave that to a separate PR :joy:

developit:
Nice.

Issue #167, opened 2020-10-20
Make npm registry url configurable


marvinhagemeister:
Just came across one of our client projects, which uses their own npm registry instead of the official one. By making it configurable via cli-flag like `--registry https://my-registry.com` we could support that use case.

I'm wondering if this is how we can make our npm plugin testable too! We could spin up a fake registry since we only ever download the tarball and use the `--registry` flag in our tests to point to our mocked server.

pajaydev:
@marvinhagemeister  @developit  I would love to do this change, I can see three idea's
1. getting it from cli-flag.
2. keeping .npmrc file and reading from it.
3. Making an entry in package.json. 

Shall we support all three ideas ?.

developit:
@pajaydev I'm not sure about a CLI flag, since folks would need to pass it everywhere they run WMR.

I actually am thinking maybe as a start we could just check for `process.env.NPM_CONFIG_REGISTRY`? It's only defined when the script is run from `npm`, but that's most cases.

chrstntdd:
Hey folks! I'm running into this same issue and would be interesting in getting this into wmr. Has there been any more progress on https://github.com/preactjs/wmr/issues/151 that is blocking this from being implemented now? 

Also, specific to my use case is that this project lives as a workspace within a mono-repo. This particular module being requested is present on disk and I have a hunch that it might be easier to resolve in the fs in these cases. Maybe mono-repo support is a whole separate discussion?

NEO-SH1W4:
Olá! 👋

Sou um desenvolvedor interessado em contribuir para o projeto. Esta issue parece perfeita para começar!

Tenho experiência com JavaScript e gostaria de trabalhar neste problema.

Podem me orientar sobre:
- Alguma documentação específica para contributors?
- Padrões de código do projeto?
- Processo de review?

Obrigado! 🙌

PR #874, opened 2021-09-19
Rewrite node module handling (npm plugin)

### LINKED ISSUE #109: Switch CommonJS transform to cjs-module-lexer

### LINKED ISSUE #135: Resolve nested "browser" field
Follow-up to #127.

Some packages use the `browser` field to rewrite import paths. This is usually done to swap out node-specific code with browser-based code. There is no resolution going on, it's straight up string replacement.

```json
{
  "name": "axios",
  "browser": {
    "./adapters/http": "./adapters/xhr"
  }
}
```

**EDIT**: `uuid` is another package that needs this:

```json
"browser": {
  "./lib/rng.js": "./lib/rng-browser.js",
  "./lib/sha1.js": "./lib/sha1-browser.js",
  "./lib/md5.js": "./lib/md5-browser.js"
},
```

### LINKED ISSUE #151: Trying to understand the npm-plugin
_I've now spent a couple of hours going through the code of our `npm-plugin`, but still don't have a good grasp on the control flow. On a high level it's easy to understand what it's supposed to do, but the devil is in the details. So this issue is more of notebook for me on my journey to figure out what is going on._

Noticed that we have multiple caches that do the same thing: Resolve a module request to the actual file content. The shape of those are:

```ts
Map<ModuleSpecifier, Map<FilePath, FileContent>>
```

Those caches only differ in that the presence of a module specifier signals that an attempt to fetch it from the network/local npm cache was already made.

There are multiple resolution strategies in the current code:

1. Check in memory cache
2. Check if it is available in local `node_modules` folder
3. Check if the tarball is available in the global npm/yarn cache folder (not implemented yet)
4. Retrieve tarball from registry
5. Bundle retrieved package and store the results in the cache of 1.

The tricky bit is that we receive cache requests in parallel. The code currently does account for that via a `whens` abstraction. It's meant to only be resolved once an item has been resolved.

To ensure that parallel requests don't trigger multiple fetches of the same tarball the code uses a `memo` function to return the same Promise object as the initiator to await on.

## Improvement ideas

- Ideally we only have one file cache and different strategies to put stuff into the cache. This hopefully simplifies retrieval access and makes the caching logic more explicit.
- Rename `whenFiles` to something more descriptive like `loadFile`. Same for `whens` which could be named `pending` or something.
- Is the `whenFiles` abstraction necessary or could we reuse a plain `Promise` with multiple listeners here?

### LINKED ISSUE #285: Packages are installed twice if used in yarn workspace
In yarn workspaces, all packages are installed in the node_modules folder of the workspace, not in each project's folder. Because of this, what I think is happening is that WMR is looking at the node_modules inside the project folder and because it doesn't find the requested packages there, it downloads them.

### LINKED ISSUE #299: Styled Components not working
I was trying to adopt WMR in a personal project but kept running into issues. I've found that it has something to do with `styled-components`, but I'm not sure what's happening.

## Error (in browser)
`Uncaught SyntaxError: The requested module '/@npm/react-is' does not provide an export named 'typeOf'`

## Steps to Reproduce
- Run `npm init wmr`
- Run `npm i styled-components`
- Import styled-components into the `index.js` file, `import styled from 'styled-components';`
- Run `npm start`, notice the page does not render and the error in the console

### LINKED ISSUE #784: WMR does not handle the "browser" field of packages correctly
**Describe the bug**

Packages making use of the `browser` field in their `package.json` to replace certain imports with others when running in a browser context throw an error.

This is not an issue if the package defines `esmodules`, `modern`, `module`, ..., but occurs in the case of the [axios](https://www.npmjs.com/package/axios) package for example because they define only the `browser` and `main` in their `package.json`.

**To Reproduce**

1. (Add the `axios` package) and import it in any file
2. Run `npm start` (or `yarn run start` if that's more your cup of tea)
3. Navigate to the page and see the error

**Expected behavior**

WMR bundles the package and handles the `browser` field correctly.

**Bug occurs with:**

- [ ] `wmr serve`

- OS: Ubuntu 20.04 through WSL2
- Browser: all
- Node Version: 14.17.2
- WMR Version: 3.4.1

**Additional context**

Link to the spec that isn't handled correctly: https://github.com/defunctzombie/package-browser-field-spec#replace-specific-files---advanced

The following code is causing the issue:
https://github.com/preactjs/wmr/blob/fd948183c78488379b4412595b176befa28a49ba/packages/wmr/src/plugins/npm-plugin/resolve.js#L10-L17

`resolve.exports`'s `legacy` (imported as `_resolveLegacyEntry` in this file) can also return a `Record<string, string | false>` due to the `browser` field, which is not handled - the code tries to call `replace` on what is possibly an object, causing it to throw an error in cases like these.

marvinhagemeister:
This PR is a rewrite of the way we handle node modules and fetch them from the npm registry.

- Add support for `browser` field in `package.json`
- Unify `development` and `production` handling into a single plugin. Previously, we used a middleware during `development` and a rollup plugin for `production`
- Prebundles dependencies to greatly reduce the number of requests. Previously, we'd fetch each file of an npm package individually.
- Improve CommonJS handling by attempting to convert it to ESM. Note that CommonJS in general has severe limitations in an ESM world and can't be served correctly.
- Auto installing of npm packages is moved to a CLI flag `--autoInstall` and is disabled by default. The auto install feature is contained in a separate plugin.
- Support for monorepo setups. We'll now follow the node resolution algorithm and traverse upwards until we find a `node_modules` folder with the package we're looking for.
- npm registry is not a global anymore, so the foundation is in place to make it either configurable or even better: read from `.npmrc `. See #167

**todos:**

- [ ] Bring back `etag` caching
- [ ] Bring back brotli compression background upgrades
- [ ] Bring back package versioning
- [ ] Fully remove previous npm plugin

Fixes #299, fixes #784, fixes #285, fixes #135, fixes #109, fixes #151


bakeiro:
Hey!

Nice PR, I need this change in order to work with wmr... maybe I can help somehow to make this PR go through?

Issue #740, opened 2021-07-08
Pick up aliases defined in compilerOptions.paths in tsconfig.json


fabiospampinato:
**Is your feature request related to a problem? Please describe.**

I use tsconfig.json for setting up aliases, those aren't picked up automatically by WMR it seems.

**Describe the solution you'd like**

They should be picked up automatically for convenience.

**Describe alternatives you've considered**

Duplicating them in a wmr.config.js file, which I wouldn't need to even have otherwise.

**Additional context**

N/A.

fabiospampinato:
It might be important to note that aliases in WMR follow a different syntax than TS' ones, this basically requires to user to learn another syntax somewhat unnecessarily. 

Also to me it's not clear at all what exactly WMR's supported syntax is.

developit:
TypeScript's aliases are relative to a project root, which makes them unusable for things that are not TypeScript. It might be possible to write a plugin that tries to emulate this, but it can never be 100% compatible since TypeScript aliases can refer to any directory on the user's machine.

fabiospampinato:
FWIW I've been using [this](https://www.npmjs.com/package/tsconfig-paths-webpack-plugin) webpack plugin for a while now, and it works great with loaders too, one can use TS aliases from inside CSS files too for example.

IMO a similar layer of compatibility could make the experience much nicer for some people.

marvinhagemeister:
The more I think about this, the more I feel like this is something that's best done as a user plugin, similar to the one for webpack. I'm a bit hesitant to put such a feature into core as this means that we take on additional maintenance of all the resolution features in TypeScript, which are arguably more complex than the ones we have.

I wish there was a universal standard for setting up aliases though. Currently every tool does have their own (although very similar) syntax for that.

developit:
@fabiosantoscode something like this should work - it's a Rollup plugin that uses the same `tsconfig-paths` module that Webpack plugin used:

https://gist.github.com/developit/091b4f551111a31f410595ee87168e7d

fabiosantoscode:
It's really good to know this is possible, but you've been mentioning the wrong person 😄

PR #875, opened 2021-09-25
(wmr) - support tsconfig-paths out of the box

JoviDeCroock:
resolves https://github.com/preactjs/wmr/issues/740

### DIFF
--- a/.changeset/honest-vans-float.md
+++ b/.changeset/honest-vans-float.md
@@ -1,0 +1,5 @@
+---
+'wmr': minor
+---
+
+Add a tsconfig-paths plugin that will check for a tsconfig and respect the paths option when present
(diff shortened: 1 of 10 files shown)


JoviDeCroock:
Currently the netlify build fails because we use `node --experimental-modules` and `tsconfig-paths` only has a CJS export, if we'd use the built wmr version we wouldn't have this issue 😅 it will be that or contributing back an esm export to `tsconfig-paths`.

CC @developit @marvinhagemeister

marvinhagemeister:
Looks like the CI is failing because we're testing against an older Node 12.x release which doesn't include the backported commonjs<->esm interop stuff that newer 12.x releases have. Now with Node 14 being the official LTS I think we're way overdue to update to the newest 12.x release in our pipeline at minimum.

JoviDeCroock:
@marvinhagemeister updated in both netlify and the github workflow

marvinhagemeister (approved):
Awesome, thanks a bunch 🙌

PR #881, opened 2021-10-06
(wmr) - await async function so it does not always evaluate to true

JoviDeCroock:


packages/wmr/src/lib/net-utils.js
```diff
@@ -30,14 +30,13 @@ export async function isPortFree(port) {
...
-	let found = false;
 	let attempts = 0;
 
 	if (typeof port === 'string') port = parseInt(port, 10);
 
 	// Limit to 20 attempts for now
-	while (!found && attempts <= 20) {
-		if (isPortFree(port)) break;
+	while (attempts <= 20) {
+		if (await isPortFree(port)) break;
```
JoviDeCroock, line 39: This is an async function and would always evaluate to true 😅 

Also removed `found` as it was never reassigned

marvinhagemeister: whoops 😬

marvinhagemeister (approved):

Issue #592, opened 2021-05-05
With WMR 2.0.0: duplicate `<script type="isodata"></script>` in prerendered HTML, due to `preact-iso` package not updated (currently `2.0.0`)


danielweck:
WMR 2.0.0 release notes:

https://github.com/preactjs/wmr/releases/tag/wmr%402.0.0

> Support injecting prerender data by returning a data key from the prerender function

https://github.com/preactjs/wmr/pull/585/files#diff-56e786dcb60ae19cef15ade284bd414d8a37f0c6534cee82bca47498797428dd

...but `preact-iso` is currently `2.0.0` which ships with:

https://github.com/preactjs/wmr/blob/6b25180b870aab137bb29f7f9cc8295d9f0067c3/packages/preact-iso/prerender.js#L49-L51

instead of:

https://github.com/preactjs/wmr/blob/00ed4e1d428177b77fe5b98abc1e48ed739cb3e0/packages/preact-iso/prerender.js#L49-L50

See published NPM package:

https://unpkg.com/browse/preact-iso@2.0.0/prerender.js

Consequence: `<script type="isodata"></script>` is generated alongside `<script type="isodata">{"foo":"bar"}</script>` when `prerender()` returns `{ data: { foo: 'bar' } }` instead of `{ data: undefined }`

Solution: publish `preact-iso` v-next :)

developit:
lol - @danielweck we're just discussing this in Slack, it's not yet decided whether `isodata` will be produced by `wmr` or `preact-iso`.

danielweck:
My local fix is to build the entire WMR project myself from the `main` branch (`yarn && yarn ci`), and to copy the generated packages from here into my project (overriding `node_modules/wmr` etc.).

danielweck:
> also only injects the script tag when a `.data` object is returned from prerender(), which causes `hydrate()` to pick the wrong root element (it uses the parent of the script tag).

Yes indeed, I immediately noticed this regression and was going to file a separate bug. I'll leave it to you then :)

JoviDeCroock:
This is fixed in latest https://www.runpkg.com/?preact-iso@2.0.3/prerender.js#49

PR #895, opened 2021-11-15
Fix/prerender hydration

rschristian:
Hydration vs rendering is [determined by whether `script[type=isodata]` can be found](https://github.com/preactjs/wmr/blob/3401a9bfa6491d25108ad68688c067a7e17d0de5/packages/preact-iso/hydrate.js#L8-L15), however, #585 limited this to only existing when returning prerender data.

I attempted to pick through the slack as [noted here](https://github.com/preactjs/wmr/issues/592#issuecomment-832852523) , though I'm still not sure I haven't missed something that would indicate this is intentional, so let me know. This just seemed like the simplest solution to bring this up with.

If we don't want `preact-iso` to return `isodata` we'll probably need some other marker to decide whether to hydrate or not?

### DIFF
--- a/packages/preact-iso/prerender.js
+++ b/packages/preact-iso/prerender.js
@@ -46,7 +46,8 @@ export default async function prerender(vnode, options) {
 	};
 
 	try {
-		const html = await render();
+		let html = await render();
+		html += `<script type="isodata"></script>`;
 		return { html, links };
 	} finally {
 		vnodeHook = null;
--- a/.changeset/friendly-lions-joke.md
+++ b/.changeset/friendly-lions-joke.md
@@ -1,0 +1,6 @@
+---
+'preact-iso': patch
+'wmr': patch
+---
+
+Fix lack of hydration when not returning prerender data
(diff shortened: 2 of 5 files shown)


rschristian:
Those test failures don't seem to by my doing

rschristian:
There's a couple of flakey tests that seem to be taking turns on failing, reran the CI a few times now with no luck.

packages/preact-iso/prerender.js
```diff
@@ -46,7 +46,8 @@ export default async function prerender(vnode, options) {
 	};
 
 	try {
-		const html = await render();
+		let html = await render();
+		html += `<script type="isodata"></script>`;
```
developit, line 50:
```suggestion
		html += `<script type="preact-iso"></script>`;
```

developit (commented):
Good catch, we never actually finished making this change.

The idea was that `isodata` should be specific to `preact-iso` (though there's not currently an API exposed for generating it).  WMR should also inject data into a script if given `result.data`, but it needs to use a different name.

rschristian:
That makes sense. Should be corrected now, assuming I'm not missing something.

CI gods finally graced me with a pass too.

developit (approved):

Issue #896, opened 2021-11-17
Prerendering not working on Node 16


Brawl345:
**Describe the bug**
When running prerender on Node 16, the pages don't get prerendered.

**To Reproduce**

1. Install Node 16
2. Clone the example project: `npm init wmr example && cd example`
3. Run `npm run build` (which in turn runs `wmr build --prerender`)

**Expected behavior**
The page should get prerendered and the resulting HTML should contain everything. This is the build output:

```
$ npm run build

> build
> wmr build --prerender

Browserslist: caniuse-lite is outdated. Please run:
npx browserslist@latest --update-db

Why you should do it regularly:
https://github.com/browserslist/browserslist#browsers-data-updating

Wrote 23Kb to disk:
  dist\index.html 460b
  dist\index.d5d2e732.js 17Kb
  dist\assets/style.e6ee4bcf.css 499b
  dist\chunks/prerender.ea3dab95.js 4.8Kb
  dist\chunks/index.fb97a71f.js 214b
```

This is the resulting `dist/index.html`:

```html
<!DOCTYPE html>
<html lang="en">
	<head>
		<meta charset="utf-8">
		<title>WMR App</title>
		<meta name="description" content="WMR App">
		<meta name="viewport" content="width=device-width,initial-scale=1">
		<link rel="icon" href="data:">
		<link rel="modulepreload" as="script" href="/index.d5d2e732.js">
		<link rel="stylesheet" href="/assets/style.e6ee4bcf.css">
	</head>
	<body>
		<script type="module" src="/index.d5d2e732.js"></script>
	</body>
</html>
```

Now the same on **Node 14**:

```$ npm run build

> @ build C:\Users\some\path\example-project
> wmr build --prerender

Browserslist: caniuse-lite is outdated. Please run:
npx browserslist@latest --update-db

Why you should do it regularly:
https://github.com/browserslist/browserslist#browsers-data-updating

Wrote 23Kb to disk:
  dist\index.html 460b
  dist\index.d5d2e732.js 17Kb
  dist\assets/style.e6ee4bcf.css 499b
  dist\chunks/prerender.ea3dab95.js 4.8Kb
  dist\chunks/index.fb97a71f.js 214b
Prerendered 3 pages:
  /
  /about [from /]
  /error [from /]
```

You can see that it prerenders. The resulting `index.html` is correct:

```html
<!DOCTYPE html>
<html lang="en">
	<head>
		<meta charset="utf-8">
		<title>WMR App</title>
[...]

Issue #893, opened 2021-10-31
Build segfaults on Node.js v16 if config contains empty middleware array


zgoda:
**Describe the bug**
If config contains empty array of middlewares, `wmr run build` ends with segfault.

**To Reproduce**

1. Initialise new project with `npm init wmr@latest myproject`, `cd` to `myproject`
2. In `wmr.config.mjs` add `middleware: []` to config object
3. `Run npm run build`

**Expected behavior**
Build finishes successfully

**Bug occurs with:**

- [ ] `wmr` or `wmr start` (development)
- [x] `wmr build` (production)
- [ ] `wmr serve`

**Desktop (please complete the following information):**

- OS: Ubuntu 20.04.3
- Browser: N/A
- Node Version: 16.13.0; NPM Version: 8.1.2
- WMR Version: 3.7.2

**Additional context**
Sometimes build finishes fine if `middleware` is `null` or `undefined` but fails at random. When set to empty array  build failed in 20/20 attempts.

rschristian:
Can you provide the error please?

I cannot reproduce, and I don't believe middlewares are used in any part of the build process.

zgoda:
This is segfault.

Config that does not segfault:

```
import { defineConfig } from 'wmr';

// Full list of options: https://wmr.dev/docs/configuration
export default defineConfig({
    /* Your configuration here */
    alias: {
        react: 'preact/compat',
        'react-dom': 'preact/compat'
    },
    middleware: null
});
```

Config that segfaults:

```
import { defineConfig } from 'wmr';

// Full list of options: https://wmr.dev/docs/configuration
export default defineConfig({
    /* Your configuration here */
    alias: {
        react: 'preact/compat',
        'react-dom': 'preact/compat'
    },
    middleware: []
});

```

rschristian:
That's really interesting. Can you share the debug info? `wmr build --debug`

zgoda:
And on Mac OS 11.6.1:

```
$ npm run build

> build
> wmr build --prerender

Browserslist: caniuse-lite is outdated. Please run:
npx browserslist@latest --update-db

Why you should do it regularly:
https://github.com/browserslist/browserslist#browsers-data-updating

Wrote 23Kb to disk:
  dist/index.html 460b
  dist/index.8a0cfff0.js 17Kb
  dist/assets/style.e6ee4bcf.css 499b
  dist/chunks/prerender.c8def485.js 4.8Kb
[...]

PR #900, opened 2022-01-08
Fix/jsx whitespace

### LINKED ISSUE #899: Line breaks between tags and text nodes becomes whitespace if within a text node
**Describe the bug**
New lines are replaced with a space character if they occur within a text node, but this is incorrect. New lines between tags and text nodes, regardless of whether they're in a text node, should be ignored.

**To Reproduce**
```jsx
<p>
    Example text{" "}
    <a
        href="https://example-link-location/"
        target="_blank"
        rel="noreferrer"
    >
        link
    </a>
    , and this is a problematic line break.
</p>
```

This is a format that tools like Prettier will generate, putting the punctuation on a new line after the end of the anchor tag. After building with WMR, however, this gets transformed into a space:

```html
<p>Example text <a href="https://example-link-location/" target="_blank" rel="noreferrer">link</a> , and this is a problematic line break.</p>
```

**Expected behavior**
New lines between tags and text nodes should be ignored, no extra whitespace.

**Bug occurs with:**

- [ ] `wmr` or `wmr start` (development)
- [X] `wmr build` (production)
- [ ] `wmr serve`

**Additional context**
Just seems to be missing logic here: https://github.com/preactjs/wmr/blob/3401a9bfa6491d25108ad68688c067a7e17d0de5/packages/wmr/src/lib/acorn-traverse.js#L880-L894

Working on a fix myself and trying to see if there are any other inconsistencies at the moment.

rschristian:
Fixes #899

Modifies JSXText creation to strip out leading and trailing whitespace if it contains newlines before then replacing any interior whitespace with a space character.

### DIFF
--- a/.changeset/late-comics-end.md
+++ b/.changeset/late-comics-end.md
@@ -1,0 +1,5 @@
+---
+'wmr': patch
+---
+
+Ensures leading newline in text nodes is not converted to whitespace
--- a/packages/wmr/src/lib/acorn-traverse.js
+++ b/packages/wmr/src/lib/acorn-traverse.js
@@ -890,8 +890,12 @@ const TYPES = {
 							// We can drop the whole matched string if it only contains
 							// whitespace characters, because that means we're at the
 							// beginning or at the end of the JSXText node
-							const replacement = /[^\s]/.test(value) ? ' ' : '';
-							value = value.replace(/\s*\n+\s*/g, replacement);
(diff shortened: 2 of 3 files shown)


rschristian:
CI failure is a segfault on prerendering, looks like a recreation of #893 and maybe #896

developit (approved):

PR #904, opened 2022-01-24
Fix/make output directory

rschristian:
Creating the output directory (if it didn't exist) happens before the user's WMR config has been merged in, which results in `dist/` always being created (regardless of what directory is set in the config file) and this can lead to build errors if the chosen out directory does not yet exist.


developit (approved):
Good catch!

PR #905, opened 2022-01-24
fix: Last, not first, non-external script in HTML should have prerender()

rschristian:
The first script tag in the file _should_ be assumed to have a `prerender()`, however, due to the loop continuing even after the first non-external script tag is found, it was actually the _last_ script tag that was used and would need to have `prerender()`.

It should be noted that [this behavior has existed for the past 9 months](https://github.com/preactjs/wmr/commit/ba2e29f2d5b3fb1a91301c1bff4f2dfba37fb335), and fixing it could potentially cause someone's setup to break. We could just flip the error message/comments instead?

https://github.com/preactjs/wmr/blob/039f79925ecab2e277d32e1f4f7341d38a222b59/packages/wmr/src/lib/prerender.js#L106

### DIFF
--- a/.changeset/pink-turtles-switch.md
+++ b/.changeset/pink-turtles-switch.md
@@ -1,0 +1,5 @@
+---
+'wmr': patch
+---
+
+First script tag in HTML should be assumed to have prerender(), not last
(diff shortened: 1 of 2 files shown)


developit:
oooof, yeah. TBH that's a solid argument in favor of flipping the error message. I can think of cases where first script is likely not the correct one (polyfills), but I can also think of cases where the last script is incorrect (analytics).

rschristian:
Will do that instead (only thought of it after I pushed, whoops).

Analytics will probably end up being external more often than not, ensuring they get skipped. So long as the error messages matches the behavior it should be an easy and quick correction anyways.

ForsakenHarmony (approved):

PR #913, opened 2022-04-04
(fix) Avoid replacing regex group token when present in html

### LINKED ISSUE #912: Prerender: regexp capturing group replaces any instance of $1 (etc.)
Ouch! When the string `$1` occurs anywhere in `headHtml` or `body` ... every instance of `$1` gets replaced by the captured group:

https://github.com/preactjs/wmr/blob/11663db6104171bcebd434ad4443e1fcb3dcea68/packages/wmr/src/lib/prerender.js#L214

https://github.com/preactjs/wmr/blob/11663db6104171bcebd434ad4443e1fcb3dcea68/packages/wmr/src/lib/prerender.js#L217

I realised this when I injected hydration data into the HTML head, luckily so far none of my pre-rendered webpages' `body` markup contained `$1` :)

cristianbote:
Fixes #912 

Switch from token based replacement to callback based function to avoid replacement of `$1` form within html and data contents at prerender time.

### DIFF
--- a/.changeset/six-pandas-look.md
+++ b/.changeset/six-pandas-look.md
@@ -1,0 +1,5 @@
+---
+'wmr': patch
+---
+
+When prerendering avoid replacing regex group token \$1 when is present in html contents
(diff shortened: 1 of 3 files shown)


developit (approved):
Good solution!

PR #915, opened 2022-04-10
Fix: TS config file using DefineConfig in a non-type="module" project

### LINKED ISSUE #914: Failed to load wmr.config.ts
**Describe the bug**
`yarn create wmr `  
rename `wmr.config.mjs` to `wmr.config.ts`
then:

```
Error: Failed to load wmr.config.ts
Error [ERR_REQUIRE_ESM]: require() of ES Module .\node_modules\wmr\index.js from .\wmr.config.js not supported.
Instead change the require of index.js in .\wmr.config.js to a dynamic import() which is available in all CommonJS modules.
TypeError [ERR_INVALID_ARG_TYPE]: The "id" argument must be of type string. Received an instance of URL

```
**Bug occurs with:**

- [x] `wmr serve`

- OS: Win 11
- Node Version: 17
- WMR Version: newest

rschristian:
Closes #914

AFAIK that `"exports"` shouldn't cause an issue for anyone?

We also don't have `.cjs` in our allowlist of config file extensions, but could add it later I suppose. Not needed now, and I don't think anyone's asked for it.

https://github.com/preactjs/wmr/blob/caf7ff840363351b0f52eabf1eff91c3e9888c04/packages/wmr/src/lib/normalize-options.js#L96


developit (approved):

developit:
fairly certain the compressed-size-action segfault is random.

rschristian:
It's pretty consistent on Node 16+. Not all the time, but 75% IME

rschristian:
Actually, this is dumb, we should just transpile the TS configs to ESM. Gonna fix that real quick.

Issue #647, opened 2021-05-27
Fresh wmr project created with create-wmr failed to build. 


mjeffryes:
**Describe the bug**
I walked through the quick start but was unable to build the resulting project. Not sure if this is specific to my machine somehow, but thought I'd better report it in case I'm not the only one.

**To Reproduce**
```
$ npm init wmr foo
$ cd foo
$ npm run build

> @ build /Volumes/Data/workspace/source/foo
> wmr build --prerender

Error: Not supported
    at eval (eval at q$e (/Volumes/Data/workspace/source/foo/node_modules/wmr/wmr.cjs:2:2528201), <anonymous>:1:7)
    at q$e (/Volumes/Data/workspace/source/foo/node_modules/wmr/wmr.cjs:2:2528208)
    at async Y$e (/Volumes/Data/workspace/source/foo/node_modules/wmr/wmr.cjs:2:2529938)

Error: Failed to load wmr.config.mjs
Error: Not supported
TypeError [ERR_INVALID_ARG_TYPE]: The "id" argument must be of type string. Received type object

    at q$e (/Volumes/Data/workspace/source/foo/node_modules/wmr/wmr.cjs:2:2528350)
```

**Expected behavior**
Should build the example project.

**Desktop (please complete the following information):**

- OS: Mac OS X 10.15.7
- Node Version: v12.14.1
- WMR Version: 2.2.0

developit:
I believe this may be an issue with ESM on Node 12.14. Can you try deleting the `wmr.config.mjs` to see if it completes as expected?

rschristian:
Probably worth noting that we do have the `"engines"` field specifying Node >= 12.17 being acceptable. Updating to at least that point would likely be a good idea to ensure you're in supported & tested territory

https://github.com/preactjs/wmr/blob/522b124a691e57ca084adc901e7ba42498aeba8f/packages/wmr/package.json#L161-L163

mjeffryes:
deleting `wmr.config.mjs` allows it to progress a little further:
```
$ npm run build

> @ build /Volumes/Data/workspace/source/foo
> wmr build --prerender


Wrote 23Kb to disk:
  dist/index.html 469b
  dist/index.587dac27.js 17Kb
  dist/assets/style.e6ee4bcf.css 499b
  dist/chunks/index.c9e2ea4c.js 214b
  dist/chunks/prerender.37a94b4d.js 4.8Kb

Error: Not supported

[...]

PR #917, opened 2022-04-11
 Fix: TS config file using DefineConfig in a non-type="module" project

### LINKED ISSUE #914: Failed to load wmr.config.ts

rschristian:
Closes #914

Rather than transpile a `wmr.config.ts` to the default module type of the project, we should instead always transpile it to ESM. No need to touch WMR's exports this way.


packages/wmr/src/lib/compile-single-module.js
```diff
@@ -74,6 +74,7 @@ export const compileSingleModule = withCache(
 			...ROLLUP_FAST_OUTPUT,
 			dir: out,
 			assetFileNames: '[name].[ext]',
+			...(typeof filename !== 'undefined' && { entryFileNames: filename }),
```
rschristian, line 77: So I attempted to reuse `assetFileNames` here, but for whatever reason, it always created an asset with an extension of `.js`, even if `.mjs` was passed to it. `entryFileNames` seems to handle this though.

I cannot rightly say I understand the difference between the two and the Rollup docs don't really help in that regard. Someone who better understands might want to look at this.

rschristian: This _seems_ like it should work, and passes the test suite, but I don't understand the difference between the two enough to know for sure whether it's acceptable.

```suggestion
			entryFileNames: filename || '[name].[ext]',
```

developit: @rschristian that suggestion makes sense - they're doing the same thing really, since `filename` didn't exist before.

developit (approved):

packages/wmr/src/lib/normalize-options.js
```diff
@@ -106,12 +106,14 @@ export async function normalizeOptions(options, mode, configWatchFiles = []) {
...
-				configFile = resolve(options.cwd, 'wmr.config.js');
+				const tmpConfig = `wmr.config.${pkg && pkg.type === 'module' ? 'js' : 'mjs'}`;
+				configFile = resolve(options.cwd, tmpConfig);
 				await compileSingleModule(file, {
 					out: resolve('.'),
+					filename: tmpConfig,
 					hmr: false,
 					rewriteNodeImports: false,
-					format: pkg && pkg.type === 'module' ? 'es' : 'commonjs'
+					format: 'es'
```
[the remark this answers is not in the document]
rschristian: I'm scared to look at the blame, as I think I remember doing it.

Not sure why I did that, really odd.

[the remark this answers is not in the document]
developit: It's a good find anyway though - I have a feeling this might have been the cause of #647 and related things.

packages/wmr/src/lib/compile-single-module.js
rschristian: Good to know!

Didn't know if there was enough of a difference between an asset and an entry, even for this tiny, single file use case that would be an issue.

PR #934, opened 2022-08-14
Add globalThis.fetch

takurinton:
I had this conversation with @developit on Twitter.
https://twitter.com/_developit/status/1558610677157269504

wmr gives an error if there is no fetch when prerendering.
To work around this, define `globalThis.fetch` in preact-iso's prerender.


takurinton:
@developit 

I am wondering if I should include the following DOMParser in the `init()` function. What do you think?

```js
globalThis.DOMParser = new (require("jsdom").JSDOM)("").window.DOMParser;
```

packages/preact-iso/prerender.js
```diff
@@ -2,13 +2,23 @@ import { h, options, cloneElement } from 'preact';
...
 options.vnode = vnode => {
 	if (old) old(vnode);
 	if (vnodeHook) vnodeHook(vnode);
 };
 
+async function init() {
+	const fs = (await eval('u=>import(u)')('fs')).promises;
+	// eslint-disable-next-line no-undef
+	globalThis.fetch = async url => {
+		const text = () => fs.readFile('dist/' + String(url).replace(/^\//, ''), 'utf-8');
```
rschristian, line 17: We can't assume the user has `dist/` as their output directory. Will need to pass the value of `--out`.

rschristian (commented):
This should be done in WMR's prerender rather than preact-iso: https://github.com/preactjs/wmr/blob/main/packages/wmr/src/lib/prerender.js

We set a couple other things on `globalThis` there, as you can see.

Edit: To expand a little, `preact-iso` isn't tied to WMR. It very well could be used in other environments. As such, if we want to fix a shortcoming in WMR specifically, we'll want to move this into WMR's prerender script.

This also allows us to pass config values, like the output directory, easily.

takurinton:
> To expand a little, preact-iso isn't tied to WMR.

Okay, I was a little mistaken. I thought this was all inclusive.

takurinton:
@rschristian 

> To expand a little, preact-iso isn't tied to WMR.

Okay, I was a little mistaken. I thought this was all inclusive.

rschristian:
Totally fair! To be honest it's in a bit of a weird spot right now. We'd like people to use it for more than just WMR, but, being buried in this monorepo, a lot of people don't even know it exists (we're working on fixing that though).

Could we add a test case for this? Super simple fetch for markdown, like you do in your blog, would be awesome! 

You can add a test fixture here: https://github.com/preactjs/wmr/tree/main/packages/wmr/test/fixtures
And a test case here: https://github.com/preactjs/wmr/blob/main/packages/wmr/test/production.test.js#L762

Just need to build & prerender the fixture then check that the prerendered HTML contains our markdown content.

takurinton:
@rschristian 
I'm new to wmr core and still don't have the full picture and completely forgot about tests.
I'll get to it right away!! Thank you!!

packages/wmr/test/production.test.js
```diff
@@ -878,6 +878,17 @@ describe('production', () => {
 				expect(instance.output.join('\n')).toMatch(/The following error was thrown during prerendering/i);
 			});
 		});
+
+		it('should support prerendered HTML with markdown using globalThis.fetch', async () => {
+			await loadFixture('prerender-markdown-content', env);
```
rschristian, line 883: nit: The focus of this test is less the markdown and instead the resource fetching. Markdown's a great & common placeholder, but not overall important to the test. What do you think of this?

```suggestion
		it('should support fetching resources from disk during prerender', async () => {
			await loadFixture('prerender-resource-fetch', env);
```

Directory name will need to be updated too if you agree. Sadly I can't make a suggestion with file paths.

rschristian (approved):

packages/wmr/test/production.test.js
takurinton: I felt that was indeed the case.
I named it markdown because I happen to use markdown files for the resources I use for testing, and I agree with you about the naming.
I will make the correction.

takurinton:
@rschristian 
fixed!!

takurinton:
@rschristian 
One test is failing, but the rest are passing.
Can this PR be merged? Also is there anything else I can do? Thank you.

rschristian:
Nearly forgot, needs a changeset.

You can run `yarn changeset` from the root of the repo to open a CLI for creating a changeset, which is a description of the change the PR provides (goes into forming the changelog). I'd probably mark this as a `minor` due to the added functionality.

I can get around to doing this later & merging too.

takurinton:
@rschristian 
https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md

takurinton:
@rschristian 
Thanks for the changeset change!
I'm at work right now and wasn't able to respond. Appreciate it!

rschristian:
Thanks!

danielweck:
> Indeed, we seem to have a flakey test. Took me a few reruns before everything passed without issue.

Possibly related?
* https://github.com/preactjs/wmr/issues/896#issuecomment-1017979928
* https://github.com/preactjs/wmr/issues/893#issuecomment-1004408408

rschristian:
@danielweck Actually no, I haven't seen segfaults in ages. Looks to be fixed on Node's side? Not sure if others share that experience or I somehow got lucky.

The flakey tests here were unrelated. HMR issues.

danielweck:
Thanks @rschristian I'll check my Node / segfault status.

PR #935, opened 2022-09-18
Feat proposal enable fetching external resources using native fetch API in Node.js version 18 later

takurinton:
I've defined globalThis.fetch at https://github.com/preactjs/wmr/pull/934, but currently wmr can't access external resources when use prerender mode.
Therefore, I came up with a plan to make fetch available natively in Node.js 18 and later.

But Node.js 18 is not LTS yet, what do you think?

### DIFF
--- a/packages/wmr/src/lib/prerender.js
+++ b/packages/wmr/src/lib/prerender.js
@@ -97,7 +97,26 @@ async function workerCode({ cwd, out, publicPath, customRoutes }) {
 	globalThis.wmr = { ssr: { head } };
 
 	// @ts-ignore
-	globalThis.fetch = async url => {
+	globalThis.fetch = async (url, options) => {
+		const pattern = /^https?:\/\/[\w/:%#\\$&\\?\\(\\)~\\.=\\+\\-]+/;
+		const isExternalUrl = pattern.test(String(url));
+		if (isExternalUrl) {
+			const nodeVersion = process.version;
+			const isNodeVersionOver18 = Number(nodeVersion.split('.')[0].slice(1)) >= 18;
+			// Use native fetch if Node.js version over 18 when `prerender`
+			if (isNodeVersionOver18) {
+				// TODO: Solve "RangeError: Maximum call stack size exceeded"
+				// MEMO: fetch is called recursively...
+				return fetch(url, options).then(res => ({
+					text: () => res.text(),
+					json: () => res.text().then(JSON.parse)
+				}));
+			}
+			throw new Error(
+				`External links are not supported in your current Node.js version ${nodeVersion} (try v18 or later).\n  ${url}`
+			);
+		}
+
 		const text = () => fs.readFile(`${out}/${String(url).replace(/^\//, '')}`, 'utf-8');
 		return { text, json: () => text().then(JSON.parse) };
 	};
--- a/packages/wmr/src/lib/prerender.js
+++ b/packages/wmr/src/lib/prerender.js
@@ -96,36 +96,19 @@ async function workerCode({ cwd, out, publicPath, customRoutes }) {
 	let head = { lang: '', title: '', elements: new Set() };
 	globalThis.wmr = { ssr: { head } };
 
-	// Define `_fetch` to avoid infinite loop in globalThis.fetch.
-	let _fetch;
-	const nodeVersion = process.version;
-	const isNodeVersionOver18 = Number(nodeVersion.split('.')[0].slice(1)) >= 18;
-	// Native fetch is only available in Node.js 18 later.
-	if (isNodeVersionOver18) {
-		_fetch = fetch;
-	}
-	// @ts-ignore
-	delete globalThis.fetch;
-	// @ts-ignore
-	globalThis.fetch = async (url, options) => {
-		const pattern = /^https?:\/\/[\w/:%#\\$&\\?\\(\\)~\\.=\\+\\-]+/;
-		const isExternalUrl = pattern.test(String(url));
-		if (isExternalUrl) {
-			// Use native fetch if Node.js version 18 later when `prerender`
-			if (isNodeVersionOver18) {
-				return _fetch(url, options).then(res => ({
-					text: () => res.text(),
-					json: () => res.text().then(JSON.parse)
-				}));
+	if (globalThis.fetch === undefined) {
+		// @ts-ignore
+		globalThis.fetch = async url => {
+			const pattern = /^\//;
+			const isLocalFile = pattern.test(String(url));
+			if (isLocalFile) {
+				const text = () => fs.readFile(`${out}/${String(url).replace(/^\//, '')}`, 'utf-8');
+				return { text, json: () => text().then(JSON.parse) };
 			}
-			throw new Error(
-				`External links are not supported in your current Node.js version ${nodeVersion} (try v18 or later).\n  ${url}`
-			);
-		}
 
-		const text = () => fs.readFile(`${out}/${String(url).replace(/^\//, '')}`, 'utf-8');
-		return { text, json: () => text().then(JSON.parse) };
-	};
+			console.warn(`fetch is not implemented in Node.js, please upgrade to Node.js 18 or above`);
+		};
+	}
 
 	// Prevent Rollup from transforming `import()` here.
 	const $import = new Function('s', 'return import(s)');
--- a/packages/wmr/src/lib/prerender.js
+++ b/packages/wmr/src/lib/prerender.js
@@ -96,18 +96,24 @@ async function workerCode({ cwd, out, publicPath, customRoutes }) {
 	let head = { lang: '', title: '', elements: new Set() };
 	globalThis.wmr = { ssr: { head } };
 
+	const _fetch = fetch;
+	// @ts-ignore
+	delete globalThis.fetch;
 	// @ts-ignore
 	globalThis.fetch = async (url, options) => {
+		let isRequested = false;
 		const pattern = /^https?:\/\/[\w/:%#\\$&\\?\\(\\)~\\.=\\+\\-]+/;
 		const isExternalUrl = pattern.test(String(url));
 		if (isExternalUrl) {
 			const nodeVersion = process.version;
 			const isNodeVersionOver18 = Number(nodeVersion.split('.')[0].slice(1)) >= 18;
+			if (isRequested) {
+				return;
+			}
 			// Use native fetch if Node.js version over 18 when `prerender`
 			if (isNodeVersionOver18) {
-				// TODO: Solve "RangeError: Maximum call stack size exceeded"
-				// MEMO: fetch is called recursively...
-				return fetch(url, options).then(res => ({
+				isRequested = true;
+				return _fetch(url, options).then(res => ({
 					text: () => res.text(),
 					json: () => res.text().then(JSON.parse)
 				}));
--- a/packages/wmr/src/lib/prerender.js
+++ b/packages/wmr/src/lib/prerender.js
@@ -96,17 +96,41 @@ async function workerCode({ cwd, out, publicPath, customRoutes }) {
 	let head = { lang: '', title: '', elements: new Set() };
 	globalThis.wmr = { ssr: { head } };
 
+	async function localFetcher(url) {
+		const text = () => fs.readFile(`${out}/${String(url).replace(/^\//, '')}`, 'utf-8');
+		return { text, json: () => text().then(JSON.parse) };
+	}
+
+	const pattern = /^\//;
 	if (globalThis.fetch === undefined) {
+		// When Node.js version under 17, native fetch API undefined
+		// So it will give a warning when the syntax to retrieve the file and the url is passed
 		// @ts-ignore
 		globalThis.fetch = async url => {
-			const pattern = /^\//;
-			const isLocalFile = pattern.test(String(url));
-			if (isLocalFile) {
-				const text = () => fs.readFile(`${out}/${String(url).replace(/^\//, '')}`, 'utf-8');
-				return { text, json: () => text().then(JSON.parse) };
+			if (pattern.test(String(url))) {
+				return localFetcher(url);
 			}
 
-			console.warn(`fetch is not implemented in Node.js, please upgrade to Node.js 18 or above`);
+			console.warn(`fetch is not defined in Node.js version under 17, please upgrade to Node.js version 18 later`);
+		};
+	} else {
+		// At that time, implement using reassignment to avoid entering an infinite loop.
+		const fetcher = fetch;
+		// @ts-ignore
+		delete globalThis.fetch;
+
+		// When Node.js version 18 later, native fetch API is defined
+		// Override with own implementation if you want to retrieve local files in Node.js 18 later
+		// ref https://github.com/preactjs/wmr/pull/935#discussion_r977168334
+		// @ts-ignore
+		globalThis.fetch = async (url, options) => {
+			if (pattern.test(String(url))) {
+				return localFetcher(url);
+			}
+
+			// When fetching an external resource, it returns the native fetch API
+			// though `fetcher` function is an alias created to avoid infinite loops
+			return fetcher(url, options);
 		};
 	}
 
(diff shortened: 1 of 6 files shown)


packages/wmr/src/lib/prerender.js
```diff
@@ -97,7 +97,26 @@ async function workerCode({ cwd, out, publicPath, customRoutes }) {
...
+			const nodeVersion = process.version;
+			const isNodeVersionOver18 = Number(nodeVersion.split('.')[0].slice(1)) >= 18;
+			// Use native fetch if Node.js version over 18 when `prerender`
+			if (isNodeVersionOver18) {
+				// TODO: Solve "RangeError: Maximum call stack size exceeded"
+				// MEMO: fetch is called recursively...
+				return fetch(url, options).then(res => ({
+					text: () => res.text(),
+					json: () => res.text().then(JSON.parse)
+				}));
```
takurinton, line 113: fetch is called recursively.
I can't come up with a best solution.

packages/wmr/test/production.test.js
```diff
@@ -889,6 +889,31 @@ describe('production', () => {
...
 		});
+
+		it('should not support fetching resources from external link prerender', async () => {
+			await loadFixture('prerender-external-resource-fetch', env);
+			instance = await runWmr(env.tmp.path, 'build', '--prerender');
+			const code = await instance.done;
+
+			const nodeVersion = process.version;
+			const isNodeVersionOver18 = Number(nodeVersion.split('.')[0].slice(1)) >= 18;
+			if (isNodeVersionOver18) {
```
takurinton, line 900: Is there a way to run tests on multiple versions?

packages/wmr/src/lib/prerender.js
takurinton: Should I use a different name instead of globalThis.fetch or something?
For examle...
```js
globalThis.f = async (url, options) => {
 ...
}
```

takurinton:
It's similar to #934, but if I merge this, do I need to create a changeset?

packages/wmr/test/production.test.js
takurinton: Oh, and two versions, 12 and 14, are running now. This felt like a no-brainer in the future.

takurinton: I'm also considering breaking down this test.

Instead of branching using `if` in one test, write a test for Node.js version 17 or less and a test for Node.js version 18 or more, and skip if the version is not applicable. I think it is also possible.
This part is one of the things I want to discuss in this PR.

packages/wmr/src/lib/prerender.js
```diff
@@ -96,8 +96,33 @@ async function workerCode({ cwd, out, publicPath, customRoutes }) {
...
+	const isNodeVersionOver18 = Number(nodeVersion.split('.')[0].slice(1)) >= 18;
+	// Native fetch is only available in Node.js 18 later.
+	if (isNodeVersionOver18) {
+		_fetch = fetch;
+	}
+	// @ts-ignore
+	delete globalThis.fetch;
 	// @ts-ignore
-	globalThis.fetch = async url => {
+	globalThis.fetch = async (url, options) => {
```
rschristian, line 110: I don't think we want to be doing this, as it breaks user polyfills. My fault as I admittedly missed it in the previous PR.

Maybe if Node < 18, use the user defined `fetch()` if it exists, else, reassign to our "polyfill". If Node >= 18, we'll need to reassign in order to inject our file reader.

What do you think?

packages/wmr/src/lib/prerender.js
```diff
@@ -96,8 +96,33 @@ async function workerCode({ cwd, out, publicPath, customRoutes }) {
...
+		const pattern = /^https?:\/\/[\w/:%#\\$&\\?\\(\\)~\\.=\\+\\-]+/;
+		const isExternalUrl = pattern.test(String(url));
+		if (isExternalUrl) {
+			// Use native fetch if Node.js version 18 later when `prerender`
+			if (isNodeVersionOver18) {
+				return _fetch(url, options).then(res => ({
+					text: () => res.text(),
+					json: () => res.text().then(JSON.parse)
+				}));
```
rschristian, line 119: We don't need to limit this to Node 18.

While indeed, `fetch()` won't work automatically, users can polyfill it in older versions of Node. We only really care about a `fetch()` existing, not what version of Node they're using (though it's a good idea to warn if it's `undefined` and their version is older than 18)

packages/wmr/test/production.test.js
rschristian: > Instead of branching using if in one test, write a test for Node.js version 17 or less and a test for Node.js version 18 or more, and skip if the version is not applicable.

I like this idea. There's slightly different behavior between the two that (ideally) we'd test.

We can definitely alter the CI config to cover newer versions of Node.

packages/wmr/src/lib/prerender.js
takurinton: Well, now that you mention it, it makes sense.
I very much agree that it is not a good idea to override our implementation by removing what the user has defined.

We felt it appropriate to leave the implementation of `fetch` itself to the user, with our concerns limited to whether `fetch` exists and whether Node.js version 18 or higher is used.

takurinton: Given the specific code, I feel it can be simplified to this point.
```js
if (globalThis.fetch === undefined) {
    // @ts-ignore
    globalThis.fetch = async url => {
		const pattern = /^\//;
		const isLocalFile = pattern.test(String(url));
		if (isLocalFile) {
			const text = () => fs.readFile(`${out}/${String(url).replace(/^\//, '')}`, 'utf-8');
			return { text, json: () => text().then(JSON.parse) };
		}
		console.warn(`fetch is not implemented in Node.js, please upgrade to Node.js 18 or above`);
	};
}
```

rschristian: Yep, I think that looks good for a Node < v18 impl.

Not sure what to do about v18 though. We either override `globalThis.fetch` to add automatic local file handling or we allow the user to override it themselves. No way to do both, I don't think. Hmm.

takurinton: I am also very confused.
I have the same implementation on my personal blog, but my motivation for creating this PR was to implement this part on the wmr side so that the user does not have to implement it!
https://github.com/takurinton/blog.takurinton.dev/blob/main/public/prerender.tsx#L9-L12

So,  
> We either override globalThis.fetch to add automatic local file handling

I feel this is a good idea. I cannot be certain that this is always the right one....

rschristian: Thought about it a bit more and I think for Node 18 we should just override `globalThis.fetch` with our own implementation.

While yes, this won't allow users to provide their own, there should be few, if any, reasons why anyone would want to do this, and I can think of none where they'd _need_ to.

takurinton:
TODO: will fix test

packages/wmr/test/production.test.js
takurinton: I completely forgot about this.
Last time I called fetch without the leading slash here, but this test does not work with the regex changes you reviewed.
https://github.com/preactjs/wmr/blob/main/packages/wmr/test/fixtures/prerender-resource-fetch/index.js#L2

I think to need to modify the test here as well. I think the comments here will be helpful for that.
https://github.com/preactjs/wmr/pull/935#discussion_r977168334

takurinton: Oops, no correction of this was necessary.

packages/wmr/src/lib/prerender.js
```diff
@@ -96,17 +96,41 @@ async function workerCode({ cwd, out, publicPath, customRoutes }) {
...
+		// @ts-ignore
+		globalThis.fetch = async (url, options) => {
+			if (pattern.test(String(url))) {
+				return localFetcher(url);
 			}
 
-			console.warn(`fetch is not implemented in Node.js, please upgrade to Node.js 18 or above`);
+			// When fetching an external resource, it returns the native fetch API
+			// though `fetcher` function is an alias created to avoid infinite loops
+			return fetcher(url, options);
```
takurinton, line 133: overrided globalThis.fetch with our own implementation.
ref https://github.com/preactjs/wmr/pull/935#discussion_r977168334

takurinton: The `fetcher` function is an alias function defined to avoid infinite loops and is equivalent to the native `fetch` function.

takurinton: When Node.js version 18 later, I addressed this by writing conditional expressions for both `fetch` is defined and it is undefined to override. 
I also wrote corresponding tests.

takurinton: Local fetching should now be automated. If you expect a different behavior from the fetch that wmr itself is overriding, you should be able to solve this by writing the following code.

```js
export async function prerender(vnode: VNode) {
    globalThis.funkyFetch = ...
}

const fetcher = typeof window === 'undefined' ? funkyFetch : fetch;
fetcher(...);
```

packages/wmr/test/fixtures/prerender-resource-fetch/index.js
```diff
@@ -1,4 +1,4 @@
 export async function prerender() {
-	const md = await fetch('content.md').then(res => res.text());
+	const md = await fetch('/content.md').then(res => res.text());
```
takurinton, line 2: I forgot to put a leading slash in the last test, so I added it.
This is because when we implemented regular expressions for paths in this PR, relative paths would cause the test to fail, and we believe that relative paths should not be used for access.

takurinton: Now all tests work correctly (currently CI is failing, but this is a problem with wmr's testing infrastructure and will succeed after a few reruns)

takurinton: I don't seem to have the authority to rerun a failed test...

takurinton:
After merging 936 into main, follow this branch as well and run the test

packages/wmr/src/lib/prerender.js
```diff
@@ -96,11 +96,43 @@ async function workerCode({ cwd, out, publicPath, customRoutes }) {
...
+		// ref https://github.com/preactjs/wmr/pull/935#discussion_r977168334
+		// @ts-ignore
+		globalThis.fetch = async (url, options) => {
+			if (pattern.test(String(url))) {
+				return localFetcher(url);
+			}
+
+			// When fetching an external resource, it returns the native fetch API
+			// though `fetcher` function is an alias created to avoid infinite loops
+			return fetcher(url, options);
```
rschristian, line 133: The comments explaining the workflow here are excellent! Really good stuff.

packages/wmr/test/production.test.js
```diff
@@ -889,6 +889,40 @@ describe('production', () => {
 			const index = await fs.readFile(indexHtml, 'utf8');
 			expect(index).toMatch(/# hello world/);
 		});
+
+		// Whether or not to run the test to fetch external resources depends on whether or not a Node.js version 18 later
+		const isNodeVersionOver18 = Number(process.version.split('.')[0].slice(1)) >= 18;
+		// Even if you are using Node.js v18 or higher, you will get the error `globalThis.fetch is undefined` here, so get the version from process.version
+		// if (globalThis.fetch === undefined)
+		if (!isNodeVersionOver18) {
```
rschristian, line 897: nit: would get rid of the inversion here and switch the following code blocks or change the name.

`isNodeVersionUnder18` would probably be more clear.

packages/wmr/test/production.test.js
```diff
@@ -889,6 +889,40 @@ describe('production', () => {
...
+			it.skip('skip should not support fetching resources from external link prerender when Node.js v18 later', () => {});
+			it('should not support fetching resources from external link prerender when Node.js befor v17', async () => {
```
rschristian, line 899:
```suggestion
			it('should not support fetching resources from external links during prerender on Node.js < v18', async () => {
```

packages/wmr/test/production.test.js
```diff
@@ -889,6 +889,40 @@ describe('production', () => {
...
+
+				await withLog(instance.output, async () => {
+					expect(code).toBe(1);
+					expect(instance.output.join('\n')).toMatch(
+						/fetch is not defined in Node.js version under 17, please upgrade to Node.js version 18 later/
+					);
+				});
+			});
+		} else {
+			it.skip('skip should not support fetching resources from external link prerender when Node.js befor v17', () => {});
```
rschristian, line 912:
```suggestion
			it.skip('skip should not support fetching resources from external links during prerender on Node.js < v18', () => {});
```

packages/wmr/test/production.test.js
```diff
@@ -889,6 +889,40 @@ describe('production', () => {
...
+			it('should not support fetching resources from external link prerender when Node.js v18 later', async () => {
```
rschristian, line 913: This seems odd? On Node 18 it shouldn't support fetching external resources? Probably just a mistake in the test name?

rschristian:
@takurinton Hmm yeah that's a weird error in the CI. Will look into it. Not sure if it's unstable or just something odd has broken.

Are you using 18 locally? Do the tests sometimes pass?

takurinton:
> Are you using 18 locally?

yes, I am using v18.9.0
```
% node -v
v18.9.0
```
>  Do the tests sometimes pass?

Only the `● production 'demo app' should serve the demo app` test fails every time. Other than that, it works sometimes.

packages/wmr/test/production.test.js
takurinton: Oops, that was my mistake! Thanks!

rschristian:
Yeah that's the really weird error. Can't find the entry module, wonder if the resolution changed or something in Node 18.

Thanks for the extra info, that help me later. We won't block your PR on that working though, as it's outside your changes here so don't worry about those failures.

takurinton:
test name fiexd!!!

> Thanks for the extra info, that help me later. We won't block your PR on that working though, as it's outside your changes here so don't worry about those failures.

Thanks!!! I'm relieved it's outside the scope of my implementation this time around!
But I'm curious about this issue and will work on it when I have time!

rschristian (approved):
Looks good to me, thanks!

rschristian:
I'm not confident this is a test issue we can ignore. Earlier I misinterpreted one of our tests as being an error when it wasn't. Haven't used this in so long I forgot what our tests were doing.

The tests pass on the main branch and they're now running with Node 18 due to your earlier PR. If you are in fact seeing a test fail every time, I think it's an issue with this PR.

takurinton:
@rschristian 
The test on the main branch is dry run because there are no changes in the actual code.
I think the way to find out if this is really a problem with this branch is to rerun the test in another PR or create a new one to see if this is the case.
What do you think?

rschristian:
Oh shoot. Yep, you're right. Sorry about that.

I'll spend some time tracking this down when I can. Would like to cut a release afterwards but wouldn't want to do that if there might be issues (related to this or not).

takurinton:
> I'll spend some time tracking this down when I can. Would like to cut a release afterwards but wouldn't want to do that if there might be issues (related to this or not).

Thank you.
I agree with that. It is very difficult to find a drop-off point.
The test was unstable to begin with, but I am afraid that this release will have an impact.

takurinton:
@rschristian 
I created another PR and ran the test on GitHub Actions and it failed in the same way.
I still can't be certain that the fetch change has no impact, but at least it appears to have no impact.

You can see log here.
https://github.com/preactjs/wmr/actions/runs/3111711765/jobs/5044311973

rschristian:
Seems pretty stable to me, see #938.

Definitely not from the fetch impl

takurinton:
@rschristian 

> Definitely not from the fetch impl

Ok!! Thanks!!

takurinton:
@rschristian 
I also feel that the fetch implementation is fine, as I wrote an exception in the `fs.rmdir` location in #937 and it stabilized!
#937 has been closed

developit (approved):

developit (approved):

PR #936, opened 2022-09-22
Fix Node.js version of GitHub Actions workflow

takurinton:
Changed the version of Node.js used for testing and releases.
Version 18 is not yet LTS, but is being added because we want to test using version 18 at https://github.com/preactjs/wmr/pull/935 (and Node.js version 18 will be LTS in the near future)

I am a bit fearful that this change will break existing mechanisms and workflows, but I will watch CI on this branch and make modifications as appropriate.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -31,7 +31,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [14.x, 16.x, 18.x]
+        node-version: [14.x, 18.x]
     steps:
       - name: Checkout
         uses: actions/checkout@v2
@@ -65,7 +65,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [14.x, 16.x, 18.x]
+        node-version: [14.x, 18.x]
     steps:
       - name: Checkout
         uses: actions/checkout@v2
@@ -96,7 +96,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [14.x, 16.x, 18.x]
+        node-version: [14.x, 18.x]
     steps:
       - name: Checkout
         uses: actions/checkout@v2
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -31,7 +31,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [12.x, 14.x]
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -129,10 +129,10 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
-      - name: Use Node.js 14.x
+      - name: Use Node.js 16.x
         uses: actions/setup-node@v1
         with:
-          node-version: 14.x
+          node-version: 16.x
```
takurinton, line 135: The lighthouse ci part specifies 16, the current LTS.

.github/workflows/release.yml
```diff
@@ -14,10 +14,10 @@ jobs:
         uses: actions/checkout@master
         with:
           fetch-depth: 0
-      - name: Setup Node.js 12.x
+      - name: Setup Node.js 16.x
         uses: actions/setup-node@master
         with:
-          node-version: 12.x
+          node-version: 16.x
```
takurinton, line 20: Like the lighthouse ci part, it specifies the current LTS.

.github/workflows/main.yml
```diff
@@ -31,7 +31,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [12.x, 14.x]
+        node-version: [14.x, 16.x, 18.x]
```
rschristian, line 34: I don't think we need 16. Let's just do 14 and 18

rschristian (approved):

.github/workflows/main.yml
takurinton: I couldn't figure out why I didn't need 16 at first glance, why?

rschristian: Why we don't need 16? Just because it shouldn't be of issue.

Generally we'd want to support the oldest LTS (14) and newest that will become an LTS (18).

16 shouldn't cause issues so it's not worth running against as having it will just increase overall CI run time.

takurinton: I see, I was trying to check all currently supported versions, but that makes sense. Thank you very much.
I will correct it.

takurinton: Fixed!! https://github.com/preactjs/wmr/pull/936/commits/58cbdf2accf064930560c7a3d9c855983c4f49e7

rschristian (approved):

JoviDeCroock (approved):

PR #937, opened 2022-09-23
Investigation of tests failing due to Node.js version 18

takurinton:
refs
#935 
#936

### DIFF
--- a/packages/wmr/test/production.test.js
+++ b/packages/wmr/test/production.test.js
@@ -470,29 +470,54 @@ describe('production', () => {
 	describe('demo app', () => {
 		it('should serve the demo app', async () => {
 			await loadFixture('../../../../examples/demo', env);
+			// In my local environment I often get errors here...
+			// Therefore, error handling is inserted here if the directory does not exist.
 			for (const d of ['dist', 'node_modules', '.cache']) {
-				await fs.rmdir(path.join(env.tmp.path, d), { recursive: true });
+				try {
+					await fs.rmdir(path.join(env.tmp.path, d), { recursive: true });
+				} catch (e) {
+					console.error(`skip rmdir. Because ${e.message}`);
+				}
 			}
 			instance = await runWmr(env.tmp.path, 'build', '--prerender');
 			const code = await instance.done;
 			const output = instance.output.join('\n');
-			console.log(output);
+
+			// comment out temporarily
+			// console.log(output);
 			if (code !== 0 || /error/i.test(output)) {
-				console.info(output);
+				// same as above, comment out temporarily
+				// console.info(output);
 			} else {
+				// same as above, comment out temporarily
 				console.info(output.match(/Wrote .+ to disk/) + '');
 			}
+
+			// This may seem slightly inconsistent with the if statement above, but I will assume it is not a problem.
 			expect(code).toBe(0);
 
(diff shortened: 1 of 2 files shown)


takurinton:
First, do an empty commit in the main branch to see if #935  is doing anything wrong.

takurinton:
Oops, no changes to the code, so it's a dry run!

takurinton:
I inserted dummy code and ran the test, but it failed with the same error as #935 .
In other words, I thought there might not be any connection with #935.
I can't be sure, though...

takurinton:
The place where it is falling down is also due to the fact that it is a `demo app` test and there is no dist directory. This seems to be a problem with the original testing infrastructure.
However, we feel that it will take a lot of work to expose and solve the underlying problems. In fact, I feel that this foundation is not well maintained.

takurinton:
I am trying to fix a few things about my demo app test failing.
Locally it fails because the .cache directory does not exist at my end, but in CI it fails because the dist directory does not exist.
And while I can just ignore the .cache directory with error handling if it doesn’t exist, I can’t proceed to the next process if I can’t get the dist directory.
Hmm...

takurinton:
Oops!!
The test that should have failed passed...
It was certainly my wish that the test would pass, but I did not wish for it to pass when it should not.😱
image

takurinton:
@rschristian 
Whether this is a fluke or the effect of the previous commit (although I just planted the log), I don't have the authority to re-run the test, so can I ask you to re-run the test instead?

packages/wmr/test/production.test.js
```diff
@@ -470,30 +470,55 @@ describe('production', () => {
...
 				console.info(output.match(/Wrote .+ to disk/) + '');
 			}
+
+			// This may seem slightly inconsistent with the if statement above, but I will assume it is not a problem.
 			expect(code).toBe(0);
 
+			// CI will probably fail due to the absence of dist in the next line.
+			// Therefore, I check here to see if the root of the demo app exists.
+			const dir = await fs.readdir(path.join(env.tmp.path, ''));
+			console.log('../dist', dir);
```
takurinton, line 502: MEMO:
This is log typo.
I originally wanted to write
```js
console.log('/path/to/demo/app/', dir);
```

takurinton:
@rschristian implemented it! Thank you! https://github.com/preactjs/wmr/pull/938

packages/wmr/test/production.test.js
```diff
@@ -470,30 +470,55 @@ describe('production', () => {
...
+			// In my local environment I often get errors here...
+			// Therefore, error handling is inserted here if the directory does not exist.
 			for (const d of ['dist', 'node_modules', '.cache']) {
-				await fs.rmdir(path.join(env.tmp.path, d), { recursive: true });
+				try {
+					await fs.rmdir(path.join(env.tmp.path, d), { recursive: true });
+				} catch (e) {
+					console.error(`skip rmdir. Because ${e.message}`);
+				}
 			}
```
takurinton, line 481: Definitely the one to fix here, I saw the same thing and did the same thing.
https://nodejs.org/api/fs.html#fspromisesrmdirpath-options

takurinton:
Closing as this can be resolved in #938

PR #940, opened 2022-10-24
create-wmr: install eslint and eslint-config-preact as dev dependencies

joebeachjoebeach:
I noticed while using `create-wmr` that when passing `--eslint`, `eslint` and `eslint-config-preact` ended up in `"dependencies"` instead of `"devDependencies"` in my `package.json`.

### Notes
- I added eslint-disable comments because the commit hook was blocking me from committing the file with the pre-existing console statements in it. I saw disable comments in other packages so I figured that's appropriate.
- I haven't been able to manually validate this. Running `yarn prepare` in the `create-wmr` package directory hangs and uses up a whole lot of my CPU. And I'm unable to directly execute `src/index.js` as it uses ES module syntax. Telling node that it's a module results in `__dirname is not defined` errors, as `__dirname` isn't available in ES modules' scope.

### DIFF
--- a/.changeset/thick-apes-relax.md
+++ b/.changeset/thick-apes-relax.md
@@ -1,0 +1,5 @@
+---
+'create-wmr': patch
+---
+
+Install eslint and eslint-config-preact as dev dependencies
--- a/packages/create-wmr/src/index.js
+++ b/packages/create-wmr/src/index.js
@@ -63,12 +63,13 @@ sade('create-wmr [dir]', true)
 
 		if (opts.eslint) {
 			spinner.start('installing eslint configuration...');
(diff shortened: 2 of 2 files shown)


rschristian:
> Running yarn prepare in the create-wmr package directory hangs and uses up a whole lot of my CPU. 

Odd. Mind sharing your environment? Don't know what would cause that, and it's seemingly find for me.

But thanks for putting this together! Really appreciate it.

rschristian (approved):

joebeachjoebeach:
@rschristian thanks for the quick turnaround!

> Mind sharing your environment?

macOS 11.6, and it has the same behavior with node 16.18.0 and 18.11.0. not sure what's causing it, but I figure it's a "me" thing and I'll dig into it

PR #951, opened 2023-01-26
add http-compression

Kikobeats:
Hello,

I backported [@polka/compression](https://github.com/lukeed/polka/pull/148) into a standalone package, so you don't need to copy/paste anymore 🙂


rschristian:
Fully agree with what Patak wrote over on Vite: https://github.com/vitejs/vite/pull/11815#issuecomment-1514375142

Adding a dependency is always something to be careful about, and the selling point for this one is... moving a little bit of code outside the project? I don't think it's particularly compelling.

Thanks for writing a PR though.

PR #953, opened 2023-02-02
[CHORE]: Update install size

GenaroIBC:
Changing install size from 2mb to 3mb, due it is closer to 3mb than 2mb


rschristian:
Thanks, but for marketing purposes, better to be over than under.

